PR #7, opened 2019-08-09
steve init

stevemao:
I failed to run unit test on my local machine. Hopefully the CI will run for me.


cis194/week1/stevemao/src/Homework.hs
```diff
@@ -0,0 +1,31 @@
+module Homework (toDigits, toDigitsRev, doubleEveryOther, sumDigits, validate) where
```
shineli1984: In Haskell, we usually align exports like this:
```Haskell
module Homework ( toDigits
                , toDigitsRev
                , doubleEveryOther
                , sumDigits
                , validate )
```

cis194/week1/stevemao/src/Homework.hs
```diff
@@ -0,0 +1,31 @@
...
+doubleOrNot :: Integer -> Int -> Integer
+doubleOrNot a b
+  | b `mod` 2 == 0 = a * 2
+  | otherwise = a
+
+doubleEveryOther :: [Integer] -> [Integer]
+doubleEveryOther = mapInd doubleOrNot
+
+sumDigits :: [Integer] -> Integer
+sumDigits = sum . (concatMap toDigits)
```
shineli1984: The normal function application in Haskell has the highest precedence and `(.)` is `infixr 9`
This means `concatMap toDigits` do not need any brackets around it.

cis194/week1/stevemao/src/Homework.hs
```diff
@@ -0,0 +1,31 @@
...
+
+divisible :: Integer -> Bool
+divisible a
```
shineli1984: In Lecture 1, there is an example given as followed:
```
isEven :: Integer -> Bool
isEven n 
  | n `mod` 2 == 0 = True
  | otherwise      = False
```
And it was asked: `This works, but it is much too complicated. Can you see why?`
I think this function has the same problem

PR #9, opened 2019-08-10
Add in class instructions

cht8687:


README.md
```diff
@@ -1,2 +1,52 @@
...
+
+Each class has average length of 2 hours. Each week, we will pick a time and place and announce it.
+
+
+
+## Materials that we used for this class
+
+
+
+Currently we are using [Yorgey's cis194 course](https://github.com/bitemyapp/learnhaskell#yorgeys-cis194-course).
```
stevemao: There are a few different versions. We should link which version we are using

README.md
```diff
@@ -1,2 +1,52 @@
 # 2019-winter-Haskell-school
+
 2019 winter Haskell school in Sydney Australia 
+
+
+
+## Class timetables and planning
+
+This winter session is initially designed to have 10 classes(might extended).
```
shineli1984: CIS194 is a 12 week course and we will extend it with advanced contents.

README.md
```diff
@@ -1,2 +1,52 @@
...
+
+This course is widely recommended by Haskell learners. 
+
+
+
+## homework and after class discussions
+
+
+
+After each week's class, participants can complete assignments from the cis194 course and submit to this repo as a [Pull Request](https://github.com/fp-works/2019-winter-Haskell-school/pulls).
```
shineli1984: Can we say `need to complete` to emphasis on the importance of homework? Or potentially a section to point out writing Haskell is the only way of learning Haskell.

README.md
cht8687: done

README.md
cht8687: updated.

README.md
cht8687: updated. 
The session to point out writing Haskell can be in a separate PR.
Could you help? I am not expert in Haskell. 😄

shineli1984: I meant something like `if you don't do your homework and expect to be a Haskell programmer in the end of this course just by coming/reading the lecture, you are dreaming.`. But you know, in a nicer way.

cht8687: done. 😄

PR #10, opened 2019-08-10
week1 wip

tienwei:
Exercises 1,2,3. Added 4.


cis194/week1/tien/homework1/src/Lib.hs
```diff
@@ -0,0 +1,30 @@
...
+  | otherwise = []
+
+toDigitsRev :: Integer -> [Integer]
+toDigitsRev num
+  | num > 0 = reverse $ toDigits num
+  | otherwise = []
+
+-- Exercise Two: --
+buildWeights :: Integer -> [Integer]
+buildWeights num = reverse $ map ((+ 1) . (`mod` 2)) [0 .. num - 1]
```
shineli1984: We can have a look at `cycle` function, I think it'll greatly reduce the code here.
In addition to above, since `cycle` will give infinite list, please think about when to use `reverse`, should we use it here or on the list we are doubling against?

cis194/week1/tien/homework1/src/Lib.hs
```diff
@@ -0,0 +1,30 @@
...
+
+doubleEveryOther :: [Integer] -> [Integer]
+doubleEveryOther arr = zipWith (*) arr $(buildWeights . toInteger . length) arr
```
shineli1984: Like the first comment this is considered better: `doubleEveryOther arr = zipWith (*) arr . buildWeights . toInteger . length  $ arr`.

cis194/week1/tien/homework1/src/Lib.hs
```diff
@@ -0,0 +1,30 @@
...
+
+-- Exercise Three: --
+sumDigits :: [Integer] -> Integer
+sumDigits arr = sum $ map sum $ map (toDigits) arr
```
shineli1984:
```suggestion
sumDigits arr = sum . map sum . map toDigits $ arr
```

Also, have a look at `concatMap` see if you can improve this line of code using it.

cis194/week1/tien/homework1/src/Lib.hs
tienwei: @shineli1984 can you please explain more on this comment? I don't quite get this `$ arr` I thought we use `$` to eval some expression but here it's only a variable `arr`. Thanks.

PR #13, opened 2019-08-12
fix doubleEveryOther

stevemao:


cis194/week1/stevemao/src/Homework.hs
```diff
@@ -21,7 +21,7 @@ doubleOrNot a b
   | otherwise = a
 
 doubleEveryOther :: [Integer] -> [Integer]
-doubleEveryOther = mapInd doubleOrNot
+doubleEveryOther = reverse . mapInd doubleOrNot . reverse
```
shineli1984: we don’t really care about the order of the digits before we sum them up.

stevemao: It says "Double the value of every second digit beginning from the *right*."
but `mapInd` is counting from the left.

shineli1984: I was hoping you can figure out that the reverse left to mapInd is not needed.

PR #17, opened 2019-08-13
week2 Exercise 1

stevemao:


cis194/week2/stevemao/LogAnalysis.hs
```diff
@@ -0,0 +1,13 @@
+{-# OPTIONS_GHC -Wall #-}
+module LogAnalysis where
+import Log
+
+parseMessage :: String -> LogMessage
+parseMessage a = case head . words $ a of
```
shineli1984: `head`, `read` and `!!` are not safe.(they are partial functions)
We can try to use the safe versions, but that will lead us to use `>>=`, which we have not covered yet.
So a better way for now is to pattern match on `a` itself. Remember that `’a’`, `’b’`, `’c’`... can be thought of constructors for type Char. And String is just `[Char]`. 
With a little help of guard on top of pattern match, it should be possible to avoid partial functions.

stevemao: @shineli1984 I'm still not sure how to convert `String` to `Int` here

cis194/week2/stevemao/LogAnalysis.hs
```diff
@@ -0,0 +1,18 @@
...
+parseInt:: String -> Int
+parseInt a = maybe 0 id (readMaybe a :: Maybe Int)
+
+-- "I 29 la la la"
+parseMessage :: String -> LogMessage
+parseMessage a = case words a of
+  ("I":x:y:_) -> LogMessage Info (parseInt x) y
+  ("W":x:y:_) -> LogMessage Warning (parseInt x) y
+  ("E":x:y:z:_) -> LogMessage (Error . parseInt $ x) (parseInt y) z
+  _ -> Unknown a
```
shineli1984: When readMaybe comes back with Nothing, we should have Unknown messageType.

cis194/week2/stevemao/LogAnalysis.hs
```diff
@@ -0,0 +1,16 @@
+{-# OPTIONS_GHC -Wall #-}
+module LogAnalysis where
+import Log
+import Text.Read
+
+parseInt:: Char -> Int
+parseInt a = maybe 0 id (readMaybe [a] :: Maybe Int)
```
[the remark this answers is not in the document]
stevemao: I'll fix it boss

cis194/week2/stevemao/LogAnalysis.hs
stevemao: I'll fix it

PR #19, opened 2019-08-14
Week2-Task1

cht8687:


cis194/week2/robert-chang/src/LogAnalysis.hs
```diff
@@ -0,0 +1,39 @@
...
+getIntFromJust s = fromJust . safeIntReader $ s 
+
+isJustValue :: String -> Bool
+isJustValue s = isJust . safeIntReader $ s
+
+parseMessage :: String -> LogMessage
+parseMessage s = case words s of 
+             ("E": level : timestamp : msg)  -> 
+               if (and [isJustValue level, isJustValue timestamp])
+                  then LogMessage (Error $ getIntFromJust $ level) (getIntFromJust timestamp) (unwords msg)
```
stevemao: `Error $ getIntFromJust $ level` can be `Error . getIntFromJust $ level` I think

cis194/week2/robert-chang/src/LogAnalysis.hs
```diff
@@ -0,0 +1,39 @@
...
+
+getIntFromJust :: String -> Int 
+getIntFromJust s = fromJust . safeIntReader $ s 
+
+isJustValue :: String -> Bool
+isJustValue s = isJust . safeIntReader $ s
+
+parseMessage :: String -> LogMessage
+parseMessage s = case words s of 
+             ("E": level : timestamp : msg)  ->
```
shineli1984: nit: We usually have 4 space indentation and 2 space for `where` clause
```Haskell
parseMessage s = case words s of 
    ("E" ....
```

cis194/week2/robert-chang/src/LogAnalysis.hs
shineli1984: Nit: align `if`, `then` and `else`
```Haskell
if ....
then ....
else ....
```

Also, when we check if something is `Just ...` then use `fromJust`, we can always use pattern match instead.
You can have nested pattern match. Or an advance trick is to use view patterns extension: https://gitlab.haskell.org/ghc/ghc/wikis/view-patterns

cis194/week2/robert-chang/src/LogAnalysis.hs
```diff
@@ -0,0 +1,39 @@
...
+                  else Unknown s
+             ("I": timestamp : msg) ->
+               if (isJustValue timestamp)
+                  then LogMessage Info (getIntFromJust timestamp) (unwords msg)
+                  else Unknown s
+             ("W": timestamp : msg) ->
+               if (isJustValue timestamp)
+                  then LogMessage Warning (getIntFromJust timestamp) (unwords msg)
+                  else Unknown s
+             __ -> Unknown s
```
shineli1984: nit: The convention is a single `_`.

cis194/week2/robert-chang/src/LogAnalysis.hs
cht8687: thanks!

cis194/week2/robert-chang/src/LogAnalysis.hs
cht8687: This is mind blowing...

PR #20, opened 2019-08-15
Week2 wip

tienwei:
Exercise one part 1.


cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -0,0 +1,47 @@
...
+    Just i  -> i
+    Nothing -> 0
+
+parseMessage :: String -> LogMessage
+parseMessage logStr =
+  case getHead . words $ logStr of
+    "I" ->
+      LogMessage
+        Info
+        (parseMaybeInt . getNthValue 2 $ logStr)
```
shineli1984: When `parseMaybeInt`gives 0, it should be a `Unknown` message. Try more pattern matches first to get the correct behaviour.

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -0,0 +1,47 @@
...
+parseMaybeInt :: Maybe Int -> Int
+parseMaybeInt maybeValue =
+  case maybeValue of
+    Just i  -> i
+    Nothing -> 0
+
+parseMessage :: String -> LogMessage
+parseMessage logStr =
+  case getHead . words $ logStr of
+    "I" ->
```
shineli1984: We can match on the first 2, 3 elements of `words logStr` here instead of just the first element. This way we can avoid using `getHead`, `getNthValue` and `getNthTailContent`.
e.g. :
```Haskell
case words logStr of
  ( "I" : ts : rest ) -> ....
  ( "W" : ts : rest ) -> ....
  ( "E" : lv : ts : rest ) -> ....
```

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -0,0 +1,47 @@
...
+module LogAnalysis where
+
+import           Log
+import           Text.Read (readMaybe)
+
+getHead :: [String] -> String
+getHead wordList =
+  case wordList of
+    x:_ -> x
+    _   -> ""
```
shineli1984: When we find ourselves coming up with default value (here `""`) for a pattern match and that default value is not necessary for the requirement, we normally have the wrong design of our code. Please see comments on `parseMessage`.

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -0,0 +1,47 @@
...
+
+getNthValue :: Int -> String -> Maybe Int
+getNthValue n str = readMaybe (getHead . drop (n-1) . words $ str) :: Maybe Int
+
+getNthTailConent :: Int -> String -> String
+getNthTailConent n str = unwords . drop n . words $ str
```
shineli1984: The value returned by `getNthValue` and `getNthTailConent` are next to each other in a string.
This is a good indication that a pattern match on specific elements of that string should do the trick.
See comment on `parseMessage`.

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -0,0 +1,47 @@
...
+
+parseMaybeInt :: Maybe Int -> Int
+parseMaybeInt maybeValue =
+  case maybeValue of
+    Just i  -> i
+    Nothing -> 0
```
shineli1984: Same here.

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -0,0 +1,26 @@
+{-# OPTIONS_GHC -Wall #-}
+
+module LogAnalysis where
+
+import           Log
+import           Text.Read (readMaybe)
+
+parseMessage :: String -> LogMessage
+parseMessage logStr =
```
shineli1984: As a cheery on top, we can have a look at view pattern extension:
https://gitlab.haskell.org/ghc/ghc/wikis/view-patterns
, which will make this more elegant. But don't worry too much, If you can't make it work.

PR #22, opened 2019-08-15
wk2 ex 1

zehua:


cis194/week2/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,23 @@
+module Exercises where
+
+import Log
+
+-- ex1
+createLogMessage :: MessageType -> String -> [String] -> LogMessage
```
sinogermany: This is a partial function here. As per @shineli1984 we need to _avoid partial function at any cost_.

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,23 @@
...
+createLogMessage mt ts msg = LogMessage mt (read ts :: TimeStamp) (unwords msg)
```
sinogermany: Consider edge cases, what if `ts` is `"a"`? Please try some of the `readXXX` functions instead of `read` to avoid runtime exceptions.

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,23 @@
...
+
+parseWords :: [String] -> LogMessage
+parseWords ("I":t:xs)   = createLogMessage Info t xs
```
sinogermany: You can get rid of `createLogMessage` and directly use `LogMessage` instead.   
If inline `unwords` bothers you, maybe you can put it in the `where` clause.   
As for `readXXX`, you can either do conventional pattern matching or use https://gitlab.haskell.org/ghc/ghc/wikis/view-patterns as @shineli1984 suggested.

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,23 @@
...
+    | otherwise                  = Unknown (unwords a)
+  where
+    errno = read e :: Int
+parseWords s            = Unknown (unwords s)
+
+parseMessage :: String -> LogMessage
+parseMessage = parseWords . words
+
+parse :: String -> [LogMessage]
+parse = fmap parseMessage . lines
```
sinogermany: @shineli1984 question: should we use the generic functor `fmap` or the equivalent instance bindings?

shineli1984: fmap is preferred. This means if you can switch the function instance to something else, it’ll still work.

cis194/week2/zehua/src/Exercises.hs
zehua: yep. fixed.

cis194/week2/zehua/src/Exercises.hs
zehua: made some changes but still hoping to keep `createLogMessage` as a util function.

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,33 @@
...
+import Log
+
+-- ex1
+parseInt :: String -> Maybe TimeStamp
+parseInt = readMaybe
+
+createLogMessage :: MessageType -> String -> [String] -> Maybe LogMessage
+createLogMessage mt ts msg =
+  case parseInt ts of
+    Just t -> Just $ LogMessage mt t (unwords msg)
```
shineli1984: nit:
```suggestion
    Just t -> Just . LogMessage mt t . unwords $ msg
```

cis194/week2/zehua/src/Exercises.hs
zehua: how do you tell from the type of the function that this is a partial function?

zehua:
ready for merging @shineli1984

cis194/week2/zehua/test/ExercisesSpec.hs
```diff
@@ -0,0 +1,59 @@
...
+        LogMessage Info 4 "Everything normal",
+        LogMessage Info 11 "Initiating self-destruct sequence",
+        LogMessage (Error 70) 3 "Way too many pickles",
+        LogMessage (Error 65) 8 "Bad pickle-flange interaction detected",
+        LogMessage Warning 5 "Flange is due for a check-up",
+        LogMessage Info 7 "Out for lunch, back in two time steps",
+        LogMessage (Error 20) 2 "Too many pickles",
+        LogMessage Info 9 "Back from lunch",
+        LogMessage (Error 99) 10 "Flange failed!"
+        ]
```
zehua: forgot to ask whether this way of formatting lists is the recommended way. @shineli1984

shineli1984: The convention is 
```Haskell
[ element1
, element2
, element3 ]
```

PR #31, opened 2019-08-17
rest of week 2 exercises

zehua:
ex 2 to 6

### DIFF
--- a/cis194/week2/zehua/src/Exercises.hs
+++ b/cis194/week2/zehua/src/Exercises.hs
@@ -62,7 +62,7 @@ whatWentWrong = fmap (fromJust . getMessage) . filter error50 . inOrder . build
 -- ex6
 -- sort events and find the events happened before a particular log message
 findEventsBefore :: String -> [LogMessage] -> [LogMessage]
-findEventsBefore event msgs = takeWhile (notseen event) . filter known . inOrder . build $ msgs
+findEventsBefore event = takeWhile (notseen event) . filter known . inOrder . build
   where
     notseen e (LogMessage _ _ m) = e /= m
     notseen _ _                  = False
--- a/cis194/week2/zehua/src/Exercises.hs
+++ b/cis194/week2/zehua/src/Exercises.hs
@@ -1,6 +1,6 @@
 module Exercises where
 
-import           Data.Maybe (catMaybes)
+import           Data.Maybe (mapMaybe)
 import           Log
 import           Text.Read  (readMaybe)
 
@@ -53,7 +53,7 @@ inOrder (Node l lm r) = inOrder l ++ [lm] ++ inOrder r
 
 -- ex5
 whatWentWrong :: [LogMessage] -> [String]
-whatWentWrong = catMaybes . fmap getMessageMaybe . filter error50 . inOrder . build
+whatWentWrong = mapMaybe getMessageMaybe . filter error50 . inOrder . build
   where
     error50 (LogMessage (Error s) _ _) = s >= 50
     error50 _                          = False
--- a/cis194/week2/zehua/src/Exercises.hs
+++ b/cis194/week2/zehua/src/Exercises.hs
@@ -1,8 +1,8 @@
 module Exercises where
 
-import           Data.Maybe
+import           Data.Maybe (fromJust)
 import           Log
-import           Text.Read
+import           Text.Read (readMaybe)
 
 -- ex1
 parseInt :: String -> Maybe TimeStamp
@@ -36,10 +36,11 @@ parse = fmap parseMessage . lines
 -- ex2
 insert :: LogMessage -> MessageTree -> MessageTree
 insert (Unknown _) t = t
-insert lm Leaf = Node Leaf lm Leaf
+insert lm Leaf       = Node Leaf lm Leaf
 insert lm@(LogMessage _ ts _ ) (Node l lm1@(LogMessage _ ts1 _) r)
   | ts < ts1 = Node (insert lm l) lm1 r
   | otherwise = Node l lm1 (insert lm r)
+insert _ t           = t
 
 -- ex3
 build :: [LogMessage] -> MessageTree
--- a/cis194/week2/zehua/src/Exercises.hs
+++ b/cis194/week2/zehua/src/Exercises.hs
@@ -1,7 +1,8 @@
 module Exercises where
 
-import Text.Read
-import Log
+import           Data.Maybe
+import           Log
+import           Text.Read
 
 -- ex1
 parseInt :: String -> Maybe TimeStamp
@@ -10,7 +11,7 @@ parseInt = readMaybe
 createLogMessage :: MessageType -> String -> [String] -> Maybe LogMessage
 createLogMessage mt ts msg =
   case parseInt ts of
-    Just t -> Just . LogMessage mt t . unwords $ msg
+    Just t  -> Just . LogMessage mt t . unwords $ msg
     Nothing -> Nothing
 
 parseWords :: [String] -> Maybe LogMessage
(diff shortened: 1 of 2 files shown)


cis194/week2/zehua/test/ExercisesSpec.hs
```diff
@@ -1,9 +1,16 @@
 module ExercisesSpec where
 
+import Data.List (sort)
 import Test.Hspec
 import Exercises
 import Log
 
+-- ad-hoc instance def to make it sortable
+instance Ord LogMessage where
```
zehua: probably not the right thing to do here 😓 just trying to be lazy in the test to not to have to manually sort the results and to assert that `inOrder . build === sort` (not exactly, need a filter to remove unknowns before `sort`).

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -31,3 +32,50 @@ parseMessage s =
...
+inOrder (Node l lm r) = inOrder l ++ [lm] ++ inOrder r
+
+-- ex5
+whatWentWrong :: [LogMessage] -> [String]
+whatWentWrong = fmap (fromJust . getMessage) . filter error50 . inOrder . build
+  where
+    error50 (LogMessage (Error s) _ _) = s >= 50
+    error50 _                          = False
+    getMessage (LogMessage _ _ m) = Just m
+    getMessage _                  = Nothing
```
zehua: don't want to use `Maybe` here because we know from the filtering with `error50` that there would be no `Unknown`. But the compiler would complain. :|

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -31,3 +32,50 @@ parseMessage s =
...
 parse :: String -> [LogMessage]
 parse = fmap parseMessage . lines
+
+-- ex2
+insert :: LogMessage -> MessageTree -> MessageTree
+insert (Unknown _) t = t
+insert lm Leaf = Node Leaf lm Leaf
+insert lm@(LogMessage _ ts _ ) (Node l lm1@(LogMessage _ ts1 _) r)
+  | ts < ts1 = Node (insert lm l) lm1 r
+  | otherwise = Node l lm1 (insert lm r)
```
zehua: compiler complains about `Unknown` being a possible value in the node, which is not really possible if the `MessageTree` is an output from `insert`. It would be a runtime error if someone passes in such a `MessageTree` as the initial value.
```
/Users/zehualiu/src/mine/2019-winter-Haskell-school/cis194/week2/zehua/src/Exercises.hs:38:1: warning: [-Wincomplete-patterns]
    Pattern match(es) are non-exhaustive
    In an equation for ‘insert’:
        Patterns not matched: (LogMessage _ _ _) (Node _ (Unknown _) _)
   |
38 | insert (Unknown _) t = t
   | ^^^^^^^^^^^^^^^^^^^^^^^^...
```

cis194/week2/zehua/test/ExercisesSpec.hs
zehua: not sure how to declare this only for the test for `ex4`, as it's only used in https://github.com/fp-works/2019-winter-Haskell-school/pull/31/files#diff-70139cd9b5806c5cd766b5ccfea9affcR125

zehua: followed the advice from the compiler to use `newtype` to do this.

Ideally, still need a way to make it visible only to the test for `ex4`.

cis194/week2/zehua/src/Exercises.hs
```diff
@@ -1,7 +1,8 @@
 module Exercises where
 
-import Text.Read
-import Log
+import           Data.Maybe
+import           Log
+import           Text.Read
```
zehua: does Haskell import order differentiate between system modules and user-defined modules? e.g., all system ones first then user-defined ones, like in PEP8 for python.

```
import           Data.Maybe
import           Text.Read

import           Log
```

cis194/week2/zehua/src/Exercises.hs
shineli1984: If we remove the pattern `insert (Unknown _) t = t` and add below as the last pattern:
```Haskell
insert _ _ -> t
```
Then we will not have this `non-exhaustive` warning. Also, I think this is a sensible logic.

A note on `non-exhaustive pattern match` warning: They should not be ignored, because it essentially says out function is partial.

cis194/week2/zehua/src/Exercises.hs
shineli1984: Folding is what you want here I think.

cis194/week2/zehua/src/Exercises.hs
shineli1984: Conventionally, yes.
Also, it's clearer these years that it's better to import `qualified` or specify functions that needed. e.g.
```Haskell
import Data.Maybe qualified as Maybe
-- or
import Data.Maybe (fromJust)
```

zehua: updated

cis194/week2/zehua/src/Exercises.hs
zehua: added `insert _ t = t`. still need `insert (Unknown _) t = t` at the beginning so that the `Leaf` case will not kick in first.

cis194/week2/zehua/src/Exercises.hs
zehua: not sure how to fold it. do you mean `concat . F.toList` or `concatMap`?

anyway, it looks like `catMaybes` works better than `fromJust`. I have changed it to use `catMaybes`.

cis194/week2/zehua/test/ExercisesSpec.hs
```diff
@@ -1,8 +1,24 @@
...
+  a1 <= a2 = let b1 = getLogMessage a1
+                 b2 = getLogMessage a2
+             in b1 `lteq` b2
+               where
+                 Unknown _ `lteq` _         = True
+                 _ `lteq` Unknown _         = False
+                 LogMessage _ ts1 _ `lteq` LogMessage _ ts2 _ = ts1 <= ts2
+
+sortLogMessages :: [LogMessage] -> [LogMessage]
+sortLogMessages = fmap getLogMessage . sort . fmap SortableLogMessage
```
zehua: Is there a way to make these only visible to the test for `ex4`?

zehua:
@shineli1984 changes in

cis194/week2/zehua/test/ExercisesSpec.hs
shineli1984: You can add a where clause in `ex4` test to have this function defined there.

zehua: moved into `sortLogMessages` into the test for ex4 with `let`. could not make `where` work there.

But the `newtype` and `instance` could not be moved there.

shineli1984: If you have 4 spaces indentation under do notation, we can try a where clause in then end with 2 space indentation like this:
```Haskell
... $ do
    ...
  where
    a = ...
```
Using let is fine here anyway.

zehua: Still could not make `where` work. Tried:

```
    it "combines with build should be equivalent to sort without Unknown" $ do
        let input  =
                [ (LogMessage Warning 124 "third")
                , (LogMessage Info 122 "first")
                , (LogMessage (Error 1) 123 "second")
                , (LogMessage (Error 2) 125 "fourth") ]
        (inOrder . build) input `shouldBe` sortLogMessages input
      where
        sortLogMessages :: [LogMessage] -> [LogMessage]
        sortLogMessages = fmap getLogMessage . sort . fmap SortableLogMessage
```

ghc reports error:
```
error: parse error on input ‘describe’
    |
143 |   describe "ex5" $ do
    |   ^^^^^^^^

--  While building package Week2-0.0.0 using:
      .stack/setup-exe-cache/x86_64-osx/Cabal-simple_mPHDZzAJ_2.4.0.1_ghc-8.6.5 --builddir=.stack-work/dist/x86_64-osx/Cabal-2.4.0.1 build lib:Week2 test:spec --ghc-options " -fdiagnostics-color=always"
    Process exited with code: ExitFailure 1
```

shineli1984: okay. Maybe try less indent on where to be 2 space less than `it`. If this doesn’t work, I need to run this on my machines to figure out why.

zehua: tried various things. as long as `where` is used, it is always the same complaint. 😞

PR #33, opened 2019-08-18
week2 -  exercie1 using view patterns + exercise 2, 3, 4, 5

tienwei:
Any feedback will be greatly appreciated. Thanks!

### DIFF
--- a/cis194/week2/tien/homework2/src/LogAnalysis.hs
+++ b/cis194/week2/tien/homework2/src/LogAnalysis.hs
@@ -1,26 +1,80 @@
 {-# OPTIONS_GHC -Wall #-}
+{-# LANGUAGE ViewPatterns #-}
 
 module LogAnalysis where
 
 import           Log
 import           Text.Read (readMaybe)
 
+-- Exercise 1 --
+-- parseMessage :: String -> LogMessage
+-- parseMessage logStr =
+--   case words logStr of
+--     ("I":ts:rest) ->
+--       case readMaybe ts :: Maybe Int of
+--         Just i -> LogMessage Info i (unwords rest)
+--         _      -> Unknown logStr
+--     ("W":ts:rest) ->
+--       case readMaybe ts :: Maybe Int of
+--         Just i -> LogMessage Warning i (unwords rest)
+--         _      -> Unknown logStr
+--     ("E":lv:ts:rest) ->
+--       case (readMaybe lv :: Maybe Int, readMaybe ts :: Maybe Int) of
+--         (Just i, Just j) -> LogMessage (Error i) j (unwords rest)
+--         _                -> Unknown logStr
+--     _ -> Unknown logStr
+-- Use ViewPatterns --
 parseMessage :: String -> LogMessage
-parseMessage logStr =
-  case words logStr of
-    ("I":ts:rest) ->
-      case readMaybe ts :: Maybe Int of
-        Just i -> LogMessage Info i (unwords rest)
-        _      -> Unknown logStr
-    ("W":ts:rest) ->
-      case readMaybe ts :: Maybe Int of
-        Just i -> LogMessage Warning i (unwords rest)
-        _      -> Unknown logStr
-    ("E":lv:ts:rest) ->
-      case (readMaybe lv :: Maybe Int, readMaybe ts :: Maybe Int) of
-        (Just i, Just j) -> LogMessage (Error i) j (unwords rest)
-        _                -> Unknown logStr
-    _ -> Unknown logStr
+parseMessage (words -> "I":(readMaybe -> Just ts):msg) =
+  LogMessage Info ts (unwords msg)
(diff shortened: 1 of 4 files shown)


cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -1,26 +1,56 @@
...
+  LogMessage (Error lvl) ts (unwords msg)
+parseMessage logStr = Unknown logStr
 
 parse :: String -> [LogMessage]
 parse = fmap parseMessage . lines
+
+-- Exercise 2 --
+constrcutTreeByTs ts1 ts2 log1 log2 left right
+  | ts1 > ts2 = Node left log2 (insert log1 right)
+  | ts1 < ts2 = Node (insert log1 left) log2 right
```
sinogermany: That will result in a non exhaustive pattern matching warning. Maybe you can either use `ts1 <= ts2` or just use `otherwise`. You may turn on `--ghc-options="-Wall -Werror"` in your `stack` command and you will be forced to fix those :)

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -1,26 +1,56 @@
...
+  constrcutTreeByTs ts1 ts2 log1 log2 left right
+insert log1@(LogMessage (Error _) ts1 _) (Node left log2@(LogMessage _ ts2 _) right) =
+  constrcutTreeByTs ts1 ts2 log1 log2 left right
+insert log1@(LogMessage _ ts1 _) (Node left log2@(LogMessage _ ts2 _) right) =
+  constrcutTreeByTs ts1 ts2 log1 log2 left right
+insert _ msgTree = msgTree
+
+-- Exercise 3 --
+build :: [LogMessage] -> MessageTree
+build = foldl (flip insert) Leaf
```
sinogermany:
```suggestion
build = foldr insert Leaf
```

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -1,26 +1,56 @@
...
-    ("W":ts:rest) ->
-      case readMaybe ts :: Maybe Int of
-        Just i -> LogMessage Warning i (unwords rest)
-        _      -> Unknown logStr
-    ("E":lv:ts:rest) ->
-      case (readMaybe lv :: Maybe Int, readMaybe ts :: Maybe Int) of
-        (Just i, Just j) -> LogMessage (Error i) j (unwords rest)
-        _                -> Unknown logStr
-    _ -> Unknown logStr
+parseMessage (words -> "I":(readMaybe -> Just ts):msg) =
```
shineli1984: This is a suggestion on code formatting:
```suggestion
parseMessage ( words -> "I" : ( readMaybe -> Just ts ) : msg ) =
```

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -1,26 +1,56 @@
...
 parse :: String -> [LogMessage]
 parse = fmap parseMessage . lines
+
+-- Exercise 2 --
+constrcutTreeByTs ts1 ts2 log1 log2 left right
+  | ts1 > ts2 = Node left log2 (insert log1 right)
+  | ts1 < ts2 = Node (insert log1 left) log2 right
+
+insert :: LogMessage -> MessageTree -> MessageTree
+insert log (id -> Leaf) = Node Leaf log Leaf
```
shineli1984: `(id -> Leaf)` is the same as pattern match on `Leaf`.
```suggestion
insert log Leaf = Node Leaf log Leaf
```

cis194/week2/tien/homework2/src/LogAnalysis.hs
```diff
@@ -1,26 +1,56 @@
...
+insert log1@(LogMessage (Error _) ts1 _) (Node left log2@(LogMessage (Error _) ts2 _) right) =
```
shineli1984: The 3 pattern matches from this line onwards express the following scenarios:
- insert `Error` to a `Node` tagged by `Error`
- insert `Error` to a `Node` tagged by any `LogMessage`
- insert any `LogMessage` into a `Node` tagged by  any `LogMessage`
All 3 above has the same logic after match: `constrcutTreeByTs ts1 ts2 log1 log2 left right`
Does this means, we only need the last one? as in, when the log being inserted and the Node that the log is inserting to both have `LogMessage` as constructor in them.

tienwei: I think I tried to get ts1 and ts2 from different log messages as there are 2 different structures of the log messages.

shineli1984: because we pattern match to get ONLY the timestamp from a message constructed by LogMessage, we can always ignore the message type.

PR #36, opened 2019-08-22
DanielM Week 2 solutions

danielmcmillan:


cis194/week2/daniel-mc/LogAnalysis.hs
```diff
@@ -0,0 +1,65 @@
...
+parseMessageType ("W" : rest) = Just (Warning, rest)
+parseMessageType ("I" : rest) = Just (Info, rest)
+parseMessageType _ = Nothing
+
+parseTsMsg :: [String] -> Maybe (TimeStamp, String)
+parseTsMsg ((safeRead -> Just ts) : rest) = Just (ts, unwords rest)
+parseTsMsg _                              = Nothing
+
+parseMessageArray :: [String] -> LogMessage
+parseMessageArray (parseMessageType -> Just (msgType, parseTsMsg -> Just (ts, msg))) =
```
shineli1984: Nit: Ususally in Haskell, we don't use `Array` inside function name becuase we can use the palura form.
`parseMessageWords`

cis194/week2/daniel-mc/LogAnalysis.hs
```diff
@@ -0,0 +1,65 @@
...
+  LogMessage msgType ts msg
+parseMessageArray msg = Unknown (unwords msg)
+
+parseMessage :: String -> LogMessage
+parseMessage = parseMessageArray . words
+
+parse :: String -> [LogMessage]
+parse x = map parseMessage (lines x)
```
shineli1984:
```Haskell 
parse x = map parseMessage (lines x)
```
is the same as
```Haskell 
parse x = map parseMessage $ lines x
```
is the same as 
```Haskell 
parse x = map parseMessage . lines $ x
```
is the same as 
```Haskell 
parse = map parseMessage . lines
```
The last one is preferred form in Haskell

cis194/week2/daniel-mc/LogAnalysis.hs
```diff
@@ -0,0 +1,65 @@
...
+
+-- Exercise 5
+
+isRelevantInfo :: LogMessage -> Bool
+isRelevantInfo (LogMessage (Error level) _ _) = level >= 50
+isRelevantInfo _                              = False
+
+getMessage :: LogMessage -> String
+getMessage (LogMessage _ _ msg) = msg
+getMessage (Unknown msg) = msg
```
shineli1984: Nit:
The second case here
``` Haskell
getMessage (Unknown msg) = msg
```
has redundent business logic, as in that it does not matter at all what's the return value when message is `Unknown`. It can be an empty string or even a random string.
This is an indication that we are trying to bolt the uasge of `map` onto `whatWentWrong` function. It's better IMO, to do folding or recurision in `whatWentWrong`, which should combine filtering and maping together to avoid this logic.

danielmcmillan: Makse sense. I had a go using foldr

danielmcmillan:
Thanks for the comments @shineli1984 !

cis194/week2/daniel-mc/LogAnalysis.hs
```diff
@@ -0,0 +1,65 @@
...
+inOrder (Node left root right) = inOrder left ++ [root] ++ inOrder right
+
+-- Exercise 5
+
+isRelevantInfo :: LogMessage -> Bool
+isRelevantInfo (LogMessage (Error level) _ _) = level >= 50
+isRelevantInfo _                              = False
+
+addMessageIfRelevant :: LogMessage -> [String] -> [String]
+addMessageIfRelevant lm@(LogMessage _ _ msg) msgs | isRelevantInfo lm = msg : msgs
```
danielmcmillan: I wonder is it fine for the guard to fall through to the next pattern like this?

shineli1984: Yes. This is exactly what we do in production code base everyday.

sinogermany: @shineli1984 @danielmcmillan you mean if we don't use `otherwise` it'll fall through to the next pattern if the guard is not satisfied?

danielmcmillan: @sinogermany Yes it seems that way, which is handy when you just want an additional condition on a pattern like in this case.

danielmcmillan:
I'm happy for this to be merged

PR #38, opened 2019-08-23
week3 ex 1 skips first attempt

zehua:
the best that I could come out with for problem 1 for now.

Will try to think about how to shorten `every`.


cis194/week3/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,28 @@
...
+-- dropWhileEnd does not seem to work on infinite list :|
+-- skips = dropWhileEnd null . map every . zip [0..] . repeat
+
+-- <*> S combinator on reader monad to implement s x y z = (x z) (y z)
+--   https://kseo.github.io/posts/2016-12-24-reader-monad-and-ski-combinators.html
+-- skips = map every . zip [0..] . (flip replicate <*> length)
+-- skips = take . length <*> map every . zip [0..] . repeat
+
+-- seems to be the most readable
+skips l = map every . zip [0..] $ replicate (length l) l
```
shineli1984: just a note here, if we flip replicate then we can make this point free. `replicate’ <*> length` is the same as `\l -> replicate (length l) l`

zehua: yeah, that's the version in line 24, end up being longer than this version by 1 char, so I opted for this one, also I assume it's more intuitive to understand with a non point free version.

shineli1984: Once we fully understand what an applicative truly is. <*> will make more sense from a whole meal programming angle. But that’s beyond week11.

cis194/week3/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,28 @@
...
+
+skips :: [a] -> [[a]]
+-- infinite list first then cut off
+-- skips = takeWhile (not . null) . map every . zip [0..] . repeat
+-- dropWhileEnd does not seem to work on infinite list :|
+-- skips = dropWhileEnd null . map every . zip [0..] . repeat
+
+-- <*> S combinator on reader monad to implement s x y z = (x z) (y z)
+--   https://kseo.github.io/posts/2016-12-24-reader-monad-and-ski-combinators.html
+-- skips = map every . zip [0..] . (flip replicate <*> length)
```
zehua: @shineli1984 is there a way to get rid of the bracket?

cis194/week3/zehua/test/ExercisesSpec.hs
```diff
@@ -0,0 +1,18 @@
...
+spec :: Spec
+spec = do
+  describe "ex1" $ do
+    it "works with empty" $ do
+      skips ([] :: [Int]) `shouldBe` []
+
+    it "works with non empty" $ do
+      skips "ABCD" `shouldBe` ["ABCD", "BD", "C", "D"]
+      skips "hello!" `shouldBe` ["hello!", "el!", "l!", "l", "o", "!"]
+      skips ([1] :: [Int])`shouldBe` [[1]]
```
zehua: have to add `:: Int` to avoid some ghc type-default warning

cis194/week3/zehua/src/Exercises.hs
zehua: agree. just need to do more of this to wrap my head around it

cis194/week3/zehua/src/Exercises.hs
shineli1984: `. ap replicate’ $ length where replicate’ = ...`

cis194/week3/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,80 @@
...
+-- using drop avoids having to check the edge cases of length <= 2
+--localMaxima []    = []
+--localMaxima [_]   = []
+--localMaxima [_,_] = []
+localMaxima =
+    -- snd3 from Data.Tuple.Utils is not from haskell-platform :|
+    map (\(_, x, _) -> x)
+    . filter (\(x, y, z) -> x < y && y > z)
+    -- $ zip3 l (drop 1 l) (drop 2 $ l) l
+    . (zip3 <$> id <*> drop 1 <*> drop 2)
```
sinogermany:
```suggestion
    . (zip3 <*> drop 1 <*> drop 2)
```
(Thanks for the cool suggestion @shineli1984 )

cis194/week3/zehua/src/Exercises.hs
```diff
@@ -1,25 +1,80 @@
...
+histogram =
+    (++ "==========\n0123456789\n")
+    . unlines
+    . dropWhile (all isSpace)
+    . transpose
+    . map sort
+    -- . (<*>) (
+    . ap (
+        -- \i -> map $ (" *"!!) . fromEnum . (==i)
+        map . (((" *"!!) . fromEnum) .) . (==)
```
shineli1984: `!!` is a partial function, we must avoid it at all cost.

cis194/week3/zehua/src/Exercises.hs
zehua: ah, this is cool. I was looking at it from a `list as applicative functor` perspective, but it looks like it could also be seen as `reader as applicative functor`!

```
zip3 l (drop 1 l) (drop 2 l) === 
(zip3 l) ((drop 1) l) ((drop 2) l) ===
((-> l) zip3) ((-> l) (drop 1) ((-> l) (drop 2)) ===
zip3 <*> drop 1 <*> drop 2
```

cis194/week3/zehua/src/Exercises.hs
zehua: updated. wasn't aware of the `bool` function. It turns out to be shorter. 😅

shineli1984:
Can I merge this? @zehua

cis194/week3/zehua/src/Exercises.hs
zehua: sorry, mixed this up with the use of `<$>` in ex3. I saw it as reader applicative, but didn't realize that `f <$> id` for reader applicative is simply `f`, because

```
<$> === fmap === .
```

PR #41, opened 2019-08-24
Daniel Deng - Week 03

sinogermany:
@shineli1984 Comparing to making the code production ready, self-describing and easy to read, I'm less interested in achieving the minimum number of characters in the code. However, if you think there's more succinct ways without sacrificing readability, or as usual when there're nits, please point them out. Cheers!


cis194/week3/daniel-deng/src/Exercise03.hs
```diff
@@ -0,0 +1,24 @@
...
+starLine :: Int -> String
+starLine nTimes = spaces (10 - nTimes) ++ stars nTimes
+
+count :: Eq a => a -> [a] -> Int
+count x = length . filter (==x)
+
+starLines :: [Integer] -> [String]
+starLines xs = filter (/= spaces 10)
+             . transpose
+             . fmap (starLine . flip count xs)
```
sinogermany: @shineli1984 Which one below is more used in production code? Or is there a third way in day-to-day code?

```suggestion
             . fmap (starLine . (`count` xs))
```

cis194/week3/daniel-deng/src/Exercise03.hs
```diff
@@ -0,0 +1,24 @@
...
+             $ [0..9]
+
+histogram :: [Integer] -> String
+histogram xs = unlines $ starLines xs ++ ["==========", "0123456789"]
```
sinogermany: Personally I find the hard-coded bottom line clearer.

cis194/week3/daniel-deng/src/Exercise03.hs
```diff
@@ -0,0 +1,24 @@
...
+stars = flip replicate '*'
+
+spaces :: Int -> String
+spaces = flip replicate ' '
+
+starLine :: Int -> String
+starLine nTimes = spaces (10 - nTimes) ++ stars nTimes
+
+count :: Eq a => a -> [a] -> Int
+count x = length . filter (==x)
```
sinogermany: @shineli1984 I find it hard to believe that there's [_nothing in hoogle_](https://hoogle.haskell.org/?hoogle=Eq%20a%20%3D%3E%20a%20-%3E%20%5Ba%5D%20-%3E%20Int) for such a trivial function?

shineli1984: is this the function? https://hackage.haskell.org/package/MissingH-1.4.2.1/docs/Data-List-Utils.html#v:countElem

sinogermany: Yep I noticed that one, not in haskell platform though. Is `MissingH` a widely used package in production code?

cis194/week3/daniel-deng/src/Exercise02.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import Data.Maybe
+
+localMax :: (Integer, Integer, Integer) -> Maybe Integer
+localMax (x, y, z)
+  | y > x && y > z = Just y
+  | otherwise      = Nothing
+
+localMaxima :: [Integer] -> [Integer]
+localMaxima xs = mapMaybe localMax $ zip3 xs (drop 1 xs) (drop 2 xs)
```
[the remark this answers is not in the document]
sinogermany:
```haskell
window n = filter ((>= n) . length) . fmap (take n) . tails
```
Thanks, that's a really cool one. Question is though, what can we do with this windowed list of lists? Do we need to pattern match array in the helper function like this?

```haskell
localMax :: [Integer] -> Maybe Integer
localMax [x, y, z] | y > x && y > z = Just y
localMax _ = Nothing

localMaxima :: [Integer] -> [Integer]
localMaxima = mapMaybe localMax . window 3
```

In production code which one tends to be used more? The uncurried version (the tuple version) is clearer to me despite the fact that we still need guard, as the signature is more restrictive. What do you think? @shineli1984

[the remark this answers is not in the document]
shineli1984: In production, we’ll use an advanced data type to embed length into List type. This is called dependent type(natural number dependent list type in this case). So that lists of different length are of different type.

In a less advanced code base, we’ll usually use a window function. But without the `filter ((>= n) . length)` part because the function receive its output will still need to match on all possible list length.

You are right that a triple is more expressive of the intend here. It’s a compromise compared to dependent type. And a list is arguably a little bit worse.

cis194/week3/daniel-deng/src/Exercise01.hs
```diff
@@ -0,0 +1,7 @@
+module CIS194.Homework03.Exercise01 (skips) where
+
+takeEvery :: Int -> [a] -> [a]
+takeEvery n = fmap snd . filter ((== n) . fst) . zip (cycle [1..n])
+
+skips :: [a] -> [[a]]
+skips = fmap . flip takeEvery <*> enumFromTo 1 . length
```
zehua: `enumFromTo 1` === `range . (1,)` with `import Data.Ix`

sinogermany: > `enumFromTo 1` === `range . (1,)` with `import Data.Ix`

Very cool! Thanks for pointing it out @zehua .

zehua: flipping takeEvery to partially apply it to the list first is a good idea. it avoids the need to generate a repeated list of the input list.

If we aim for shortening the code, we could define takeEvery as the flipped version to get rid of the flip call (4 chars) by adding two `l`s in the definition, giving us a gain of two fewet chars.  :D

cis194/week3/daniel-deng/src/Exercise02.hs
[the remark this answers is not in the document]
zehua: @shineli1984 any reference to morediscussions on dependent type?

PR #42, opened 2019-08-24
steve exercise 2

stevemao:


stevemao:
Unfortunately this solution doesn't seem to be as simple as a for loop :( (assume we won't change the requirement anymore in the future.)
Would be good to see a solution that's as easy as for loop and still readable and maintainable.

cis194/week3/stevemao/Golf.hs
```diff
@@ -0,0 +1,21 @@
...
+removeFirst :: [a] -> [a]
+removeFirst (_: xs) = xs
+
+getSecondIfMax :: (Integer, Integer, Integer) -> [Integer] -> [Integer]
+getSecondIfMax (a, b, c) list
+  | b > a && b > c = [b] ++ list
+  | otherwise = list
+
+localMaxima :: [Integer] -> [Integer]
+localMaxima list = foldr getSecondIfMax [] (zip3 list (removeFirst list) (removeFirst . removeFirst $ list))
```
sinogermany:
```suggestion
localMaxima list = foldr getSecondIfMax [] $ zip3 list (drop 1 list) (drop 2 list)
```

sinogermany:
[cis194/week3/stevemao/Main](https://github.com/fp-works/2019-winter-Haskell-school/pull/42/files#diff-2ba0860ed37ec58d692bf58585b06112) can be git ignored

cis194/week3/stevemao/Golf.hs
stevemao: Thanks @sinogermany

PR #43, opened 2019-08-25
Tien Week 3 exercises1

tienwei:
Not sure the solution is good enough. Any feedback is welcome. Thanks!


cis194/week3/tien/homework3/src/Golf.hs
```diff
@@ -0,0 +1,17 @@
...
+-- Exercise 1 --
+everyNth :: Integral b => b -> [(a, b)] -> [a]
+everyNth n = fst . unzip . filter ((==0).(`mod` n).snd)
+
+skips::[a] -> [[a]]
+skips []  = []
+skips arr = do
+  let indexArr = [1..length arr]
+  let pairs = zip arr indexArr
+  map (\x -> everyNth x pairs ) indexArr
```
shineli1984: Nit: With `where`, `do` notation is not needed, (`do` is imperative, we use it when we can't think about an abstraction to express our intention as compositions).
Here we can use:
```Haskell
    map (\x -> everyNth x pairs ) indexArr
  where 
    indexArr = [1..length arr]
    pairs = zip arr indexArr
```

shineli1984: Also, if we flip everyNth, the lambda expression on line 12 can be easily point free.
```Haskell
map (`everyNth` pairs) indexArr`
```

cis194/week3/tien/homework3/src/Golf.hs
```diff
@@ -0,0 +1,17 @@
+module Golf where
+
+-- Exercise 1 --
+everyNth :: Integral b => b -> [(a, b)] -> [a]
+everyNth n = fst . unzip . filter ((==0).(`mod` n).snd)
```
shineli1984: Nit: 
```suggestion
everyNth n = fst . unzip . filter ( ( ==0 ) . ( `mod` n ) . snd )
```
We usually put space everywhere :D

PR #44, opened 2019-08-25
improved ex1 wk3 with list applicative

zehua:
slightly shorter now


cis194/week3/zehua/src/Exercises.hs
```diff
@@ -47,10 +50,75 @@ skips :: [a] -> [[a]]
...
+--skips l = map every . zip [0..] $ replicate (length l) l
+
+-- utilizing list applicative to save calls to repeat
+-- and use a curried version of every to avoid zip
+-- everY = curry every
+-- 66 chars
+everY :: Int -> [a] -> [a]
+everY n xs = case drop n xs of
+              (y:ys) -> y : everY n ys
+              []     -> []
```
shineli1984: You can use `_` instead of `[]`

PR #45, opened 2019-08-26
Wei Ning - CIS194 Week3 Homework

powergun:
Thanks Shine for teaching!!

To run the exercises:
```
stack test
```


cis194/week3/weining/src/Golf.hs
```diff
@@ -0,0 +1,74 @@
...
+a b | c d |      (dropping a, b)
+drop  ^ ^   drop
+
+Then it uses a filter lambda to discard those that do not pass
+the local-maxima test.
+-}
+localMaxima :: [Integer] -> [Integer]
+localMaxima xs = map (!! 1) $ filter (\[a, b, c] -> b > a && b > c) triplets
+  where
+    triplets = filter (\n -> length n == 3) $ transpose [xs, drop 1 xs, drop 2 xs]
```
shineli1984: Nit (use composition style instead of application style):
```suggestion
    triplets = filter ( (== 3) . length ) . transpose $ [xs, drop 1 xs, drop 2 xs]
```

cis194/week3/weining/src/Golf.hs
```diff
@@ -0,0 +1,74 @@
...
+    | a b | c d
+  a | b c | d    (dropping a)
+a b | c d |      (dropping a, b)
+drop  ^ ^   drop
+
+Then it uses a filter lambda to discard those that do not pass
+the local-maxima test.
+-}
+localMaxima :: [Integer] -> [Integer]
+localMaxima xs = map (!! 1) $ filter (\[a, b, c] -> b > a && b > c) triplets
```
shineli1984: `!!` is a partial function. We need to avoid it at all cost.

cis194/week3/weining/src/Golf.hs
```diff
@@ -0,0 +1,74 @@
...
+n being the number of occurrence of a certain number between 0 to 9.
+> short bars are padded with space ' ' character to ensure that
+they have the same height
+
+lastly, each bar is stored in a string which is in turn stored in
+  a list; the entire histogram is generated by transposing this
+  list ("rotating" the bar chart)
+-}
+histogram :: [Integer] -> String
+histogram xs = unlines $ transpose $ map (\n -> space n ++ bar n ++ base n) columns
```
shineli1984:
```suggestion
histogram xs = unlines . transpose . fmap (\n -> space n ++ bar n ++ base n) $ columns
```
The standard trick to do above is to
- change all `$` to `.`
- add a `$` before the last value.

cis194/week3/weining/src/Golf.hs
```diff
@@ -0,0 +1,74 @@
...
+  where
+    columns = [0..9]
+    height = maximum $ map (\n -> length [x | x <- xs, x == n]) columns
```
shineli1984: You can give it a go here to use `.` and `$`.

cis194/week4/weining/src/Exercise1.hs
```diff
@@ -0,0 +1,34 @@
...
+
+fun2Original :: Integer -> Integer
+fun2Original 1 = 0
+fun2Original n | even n    = n + fun2Original (n `div` 2)
+               | otherwise = fun2Original (3 * n + 1)
+
+fun2 n = sum $
+         filter even $
+         takeWhile (> 1) $
+         iterate (bool <$> ((+ 1) . (* 3)) <*> (flip div 2) <*> even) n
```
shineli1984: nit:
```Haskell
fun2 = sum
     . filter even
     . takeWhile (> 1)
     . iterate (bool <$> ((+ 1) . (* 3)) <*> (flip div 2) <*> even)
```
To turn function applications to function composition, the standard trick is to replace `$` with `.` then add `$` before the last variable. Because the last variable is the same as the only function input to fun2, we can make this point free like above.

cis194/week4/weining/src/Exercise1.hs
```diff
@@ -0,0 +1,34 @@
...
+
+fun1Original :: [Integer] -> Integer
+fun1Original [] = 1
+fun1Original (x:xs)
+    | even x = (x - 2) * fun1Original xs
+    | otherwise = fun1Original xs
+
+fun1 :: [Integer] -> Integer
+fun1 = (foldr (*) 1) .
+       (map (bool <$> (const 1) <*> (subtract 2) <*> even))
```
shineli1984: In Haskell we often align symbols. 
```Haskell
fun1 = foldr (*) 1 -- since function application has the highest precedence, we don't need brackets around this.
     . map calc
   where
     calc = bool <$> const 1 <*> subtract 2 <*> even -- we are trying to get rid of some brackets here to reduce noise in the code base
```

cis194/week3/weining/src/Golf.hs
powergun: Thanks for the tips! Makes it a lot easier for me to remember the idiom.

shineli1984:
I think this is covered by a later PR. Hence, close

PR #48, opened 2019-08-26
Task1

cht8687:
Task1


cis194/week3/robert-change/Golf/src/Golf.hs
```diff
@@ -0,0 +1,18 @@
...
+-- skips :: [a] -> [[a]]
+-- skips l = (flip sieve) l <$> [1..length l]
+-- however the output is:
+--  skips 'ABCD'
+--  expected: ["ABCD","BD","C","D"]
+--  but got: ["ABCD","AC","AD","A"]
+
+-- attempt 2 
+skips :: [a] -> [[a]]
+skips l = (flip $ (.) <$> sieve <*> drop . (-1 +)) l <$> [1..length l]
```
shineli1984: Does the below work?
```suggestion
skips l = ( (.) <$> drop . (-1 +) <*> sieve ) l <$> [1..length l]
```

cht8687: @shineli1984 
 • Couldn't match type ‘[a]’ with ‘Int’
      Expected type: [a] -> Int -> [a]
        Actual type: Int -> [a] -> [a]
    • The function ‘(.) <$> drop . (- 1 +) <*> sieve’
      is applied to one argument,
      but its type ‘Int -> [a] -> [a]’ has only two
      In the first argument of ‘(<$>)’, namely
        ‘((.) <$> drop . (- 1 +) <*> sieve) l’
      In the expression:
        ((.) <$> drop . (- 1 +) <*> sieve) l <$> [1 .. length l]
    • Relevant bindings include
        l :: [a] (bound at src/Golf.hs:18:7)
        skips :: [a] -> [[a]] (bound at src/Golf.hs:18:1)
   |
18 | skips l = ( (.) <$> drop . (-1 +) <*> sieve ) l <$> [1..length l]
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
make: *** [run_test] Error 1

shineli1984: oh... right. nvm

PR #50, opened 2019-08-28
Task2 

cht8687:


cis194/week3/robert-change/Golf/src/Golf.hs
```diff
@@ -14,5 +14,24 @@ import Data.List.HT
...
+-- applied n => sieve n (drop (n-1) l) to each element
+-- drop (n-1) l will remove the first n-1 elements
+-- then sieve n newElement will move the index to the current n
+-- so that attempt 1 issue will be resolved 
+skips l = (\n -> sieve n (drop (n-1) l)) <$> [1..length l]
+
+
+-- using recurison 
+localMaxima :: [Integer] -> [Integer]
+localMaxima (x:xs:y:z)
```
shineli1984: nit:
In the production code base, we `xs` usually means `x`s. -- multiple x.
So here I suggests that 
```suggestion
localMaxima ( x : y : z : as )
```

cht8687: @shineli1984 what is the convention of the 4th parameter? `as` ?

shineli1984: Since we got `x, y, z` I won't be fair to have a `xs` or `ys`. `zs` may be a good choice here but `as` is also acceptable.

PR #52, opened 2019-08-29
wk4 ex1

zehua:
ex1 first


cis194/week4/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,27 @@
...
+
+-- ex1
+fun1 :: [Integer] -> Integer
+fun1 [] = 1
+fun1 (x:xs)
+  | even x = (x - 2) * fun1 xs
+  | otherwise = fun1 xs
+
+fun1' :: [Integer] -> Integer
+fun1' = foldr ((*) . (-2 +)) 1 . filter even
```
sinogermany:
```suggestion
fun1' = product . (subtract 2) . filter even
```

zehua: do you mean `map (subtract 2)`

sinogermany: > do you mean `map (subtract 2)`

Yep, `product . fmap (subtract 2) . filter even`.
Sorry for the confusion...

cis194/week4/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,27 @@
...
+
+-- ex1
+fun1 :: [Integer] -> Integer
+fun1 [] = 1
+fun1 (x:xs)
+  | even x = (x - 2) * fun1 xs
+  | otherwise = fun1 xs
+
+fun1' :: [Integer] -> Integer
+fun1' = product . fmap (subtract 2) . filter even
```
shineli1984: Just FYI here, because `-` carries 2 meanings. we need to write `((-) 2)` here for it to express subtracting.

cis194/week4/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,27 @@
...
+fun2 :: Integer -> Integer
+fun2 1 = 0
+fun2 n | even n = n + fun2 (n `div` 2)
+       | otherwise = fun2 (3 * n + 1)
+
+fun2' :: Integer -> Integer
+fun2' = foldr (+) 0
+      . filter even
+      . takeWhile (>1) -- >1 instead of /=1 to defend against negative numbers
+      . iterate (bool <$> oddv <*> evenv <*> even)
```
shineli1984: We can also write the below, you are the judge to say which one is better
```
      . iterate (bool . oddv <*> evenv <*> even)
```

sinogermany: @shineli1984 from what you said a generic `fmap` / `<$>` is better than the binding?

```haskell
instance Functor ((->) r) where
    fmap = (.)
```

So I guess `<$>` is better in this case?

shineli1984: `... <$> ... <*> ... <*>` is a very common pattern when using applicative. 
But the choice of using `.` or `<$>` for function composition is open, at least from my point of view. Because one can argue that when apply a value on `(bool <$> oddv <*> evenv <*> even)` that value will go to `oddv` first instead of `bool` as if we are adding redundant brackets like this: `(bool <$> oddv) <*> evenv <*> even`. This is to express composition first instead of `fmap`. They have different intent but same result. So either is fine.

zehua:
@shineli1984 feel free to merge this one.

cis194/week4/zehua/test/ExercisesSpec.hs
```diff
@@ -0,0 +1,31 @@
...
+import           Test.Hspec
+
+spec :: Spec
+spec = do
+  describe "ex1.fun1" $ do
+    it "works with empty" $ do
+      fun1' [] `shouldBe` fun1 []
+
+    it "works with non-empty" $ do
+      mapM_ (liftM2 shouldBe fun1' fun1)
```
zehua: is this the right way to do it (via `mapM_`)?

Also I found no difference between `liftA2` and `liftM2` here.

shineli1984: Yeah this is correct.

If the Applicative instance is defined using monad’s >>=, then liftA2 and liftM2 are the same.

PR #53, opened 2019-08-29
Daniel Deng Week 04

sinogermany:


cis194/week4/daniel-deng/src/Exercise01.hs
```diff
@@ -0,0 +1,13 @@
+module CIS194.Homework04.Exercise01 (fun1, fun2) where
+
+import Data.Bool
+
+fun1 ::  [Integer] -> Integer
+fun1 = product . fmap (subtract 2) . filter even
+
+fun2 :: Integer -> Integer
+fun2 = bool <$> handleN <*> const 0 <*> (<=1)
```
shineli1984: FYI. `bool <$> handleN` is the same as `bool . handleN`

cis194/week4/daniel-deng/src/Exercise03.hs
```diff
@@ -0,0 +1,15 @@
...
+
+map' :: (a -> b) -> [a] -> [b]
+map' f = foldr ((:) . f) []
+
+{-
+foldr f z [x1, x2, ..., xn] == x1 ‘f‘ (x2 ‘f‘ ... (xn ‘f‘ z)...)
+foldl f z [x1, x2, ..., xn] == (...((z ‘f‘ x1) ‘f‘ x2) ‘f‘...) ‘f‘ xn
+
+-- still thinking about how to implement foldl with foldr
+myFoldl :: (a -> b -> a) -> a -> [b] -> a
```
shineli1984: Nested function calls, as in one function calls another function calls another function ..., form a stack. And stack is FILO by nature. In another word, the last function being called returns first. Have this in mind, can we accumulating functions using `foldl` so when we invoke these functions, it will return as if we are folding from the right?

cis194/week4/daniel-deng/src/Exercise01.hs
```diff
@@ -0,0 +1,13 @@
...
+     where
+       handleN    = bool <$> handleOdd <*> handleEven <*> even
+       handleOdd  = fun2 . (1+) . (3*)
+       handleEven = (+) <*> fun2 . (`div` 2)
```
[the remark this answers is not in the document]
sinogermany: > `handleOdd` and `handleEven` are recursive on `fun2`. Can use utilise `iterate` instead?

@shineli1984 yes we certainly can. My question is, would we prefer wholemeal programming in general and try to avoid recursive function calls? In the production codebase, for this particular instance, do we prefer to have `iterate` and `takeWhile`?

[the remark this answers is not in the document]
shineli1984: They are both fine in terms of `wholemeal`. But most recursions (as in recursion with a lot of operational concerns around it) are not.

cis194/week4/daniel-deng/src/Exercise01.hs
```diff
@@ -0,0 +1,22 @@
...
+import Data.Bool
+
+fun1 ::  [Integer] -> Integer
+fun1 = product . fmap (subtract 2) . filter even
+
+fun2 :: Integer -> Integer
+fun2 = sum
+     . filter even
+     . takeWhile (>1)
+     . iterate (bool <$> handleOdd <*> handleEven <*> even)
```
sinogermany: @shineli1984 I think I'll still use applicative style here just to develop my muscle memory 😂

shineli1984: That's fine. Your call.

cis194/week4/daniel-deng/src/Exercise03.hs
sinogermany: > Nested function calls, as in one function calls another function calls another function ..., form a stack. And stack is FILO by nature. In another word, the last function being called returns first. Have this in mind, can we accumulating functions using `foldl` so when we invoke these functions, it will return as if we are folding from the right?

Sounds like we are implementing a runtime of a scripting language 😂

PR #54, opened 2019-08-30
week4 ex 2,3,4

zehua:


cis194/week4/zehua/test/ExercisesSpec.hs
```diff
@@ -29,3 +48,59 @@ spec = do
...
+        , "ABCDEFGHIJKLMNOPQRSTUVWXYZ" ]
+
+  describe "ex3" $ do
+    it "xor" $ do
+      xor [] `shouldBe` False
+      xor [True] `shouldBe` True
+      xor [True, False, True] `shouldBe` False
+      xor [True, False, True, False] `shouldBe` False
+      xor [True, False, True, False, False] `shouldBe` False
+      xor [True, False, True, True, False, False] `shouldBe` True
```
zehua: anyway to lift this into something to accept
```
[ (False, [])
, (True, [True])
, ... ]
```

cis194/week4/zehua/src/Exercises.hs
```diff
@@ -25,3 +26,64 @@ fun2' = foldr (+) 0
...
+xor :: [Bool] -> Bool
+xor = foldr xor2 False
+
+
+map' :: (a -> b) -> [a] -> [b]
+map' f = foldr ((:) . f) []
+
+
+myFoldl :: (a -> b -> a) -> a -> [b] -> a
+myFoldl f base = foldr (flip f) base . reverse
```
shineli1984: There are 2 folding happen here, one is `reverse` one is `foldr`. Can we use only 1 folding?
hint: What about function as accumulator for foldr? because nested function calls when evaluated is in the FILO order.

cis194/week4/zehua/test/ExercisesSpec.hs
shineli1984: try to use `mapM_` and `uncurry` on the `shouldBe`

cis194/week4/zehua/src/Exercises.hs
zehua: thanks for the suggestion. updated the code

cis194/week4/zehua/test/ExercisesSpec.hs
zehua: `uncurry` with `shouldBe` seems not that intuitive to understand. resorting to lambda expression.

zehua:
@shineli1984 PR updated

cis194/week4/zehua/src/Exercises.hs
```diff
@@ -25,3 +26,75 @@ fun2' = foldr (+) 0
...
+
+myFoldl :: (a -> b -> a) -> a -> [b] -> a
+-- myFoldl f base = foldr (flip f) base . reverse
+-- single foldr, with function as accumulator
+-- reverse' :: [a] -> [a]
+-- reverse' l = foldr (\x g -> g . (x:)) id l []
+-- reverse' l = foldr (\x g -> g . ((:) x)) id l []
+-- myFoldl f base l = foldr (flip f) base reverse'
+-- combining the two foldr into one
+-- myFoldl f base l = foldr (\x g v -> g (f v x)) id l base
```
zehua: this is probably the version that's the easiest to understand 😅

PR #55, opened 2019-08-30
Week 3 Ex 3 + Week 4 Ex 1

tienwei:
Thanks.


cis194/week3/tien/homework3/src/Golf.hs
```diff
@@ -17,3 +20,40 @@ localMaxima (x:y:z:xs)
...
+
+formGroups :: [Integer] -> [Integer]
+formGroups =
+  concat .
+  drop 1 .
+  transpose .
+  map (\(x, y) -> [x, y]) . assocs . accumArray (+) 0 (0, 9) . formPairs
+
+mapToHistogram :: [Integer] -> [String]
+mapToHistogram =
```
sinogermany: Not sure how many chars there would be - but FYI, you can use applicative style as an alternative to `if ... then ... else`:

```haskell
import Data.Bool

-- ...

mapToHistogram :: [Integer] -> [String]
mapToHistogram = map intToString
               where
                 intToString = bool <$> const " " <*> const "*" <*> (>0)
```

tienwei: Thanks @sinogermany. Will check and update. :)

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -0,0 +1,30 @@
...
+-- Exercise 1 --
+-- 1.1 --
+fun1 :: [Integer] -> Integer
+fun1 [] = 1
+fun1 (x:xs)
+  | even x = (x - 2) * fun1 xs
+  | otherwise = fun1 xs
+
+fun1' :: [Integer] -> Integer
+fun1' = product . map (subtract 2) . filter (even)
```
shineli1984: We normally don't use brackets when it's not necessary to reduce the characters to be read.
```suggestion
fun1' = product . map (subtract 2) . filter even
```

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -0,0 +1,30 @@
...
+fun2' :: Integer -> Integer
+fun2' =
+  sum .
+  filter even .
+  takeWhile (> 1) .
+  iterate
+    (\x ->
+       if even x
+         then x `div` 2
+         else 3 * x + 1)
```
shineli1984: In Haskell, we usually put symbols at the beginning of a line so a more conventional style is.
```Haskell
fun2' 
    = sum 
    . filter even 
    . takeWhile (> 1)
    . iterate (\x ->
        if even x
        then x `div` 2 -- also align then & else with if
        else 3 * x + 1)
```

tienwei: Ha this is auto-formatted by the plugin. Maybe i will try if I can config the alignment. Thanks.

cis194/week3/tien/homework3/src/Golf.hs
tienwei: @sinogermany updated the style. Thanks for the tip. It looks much neater!

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -19,12 +21,12 @@ fun2 n
...
-  filter even .
-  takeWhile (> 1) .
-  iterate
-    (\x ->
-       if even x
-         then x `div` 2
-         else 3 * x + 1)
+fun2' = sum . filter even . takeWhile (> 1) . iterate configInputs
+  where
+    configInputs = bool <$> (+ 1) . (* 3) <*> flip div 2 <*> even
```
shineli1984: Nice one here. just FYI, the `<$>` here is the same as `.`. So the above line will be the same as:
`configInputs = bool . (+ 1) . (* 3) <*> flip div 2 <*> even`

PR #56, opened 2019-08-31
[Week4]: Task1

cht8687:


cis194/week4/robert-chang/src/Lib.hs
```diff
@@ -0,0 +1,29 @@
...
+ | otherwise = fun2 (3 * n + 1)
+
+fun2' :: Integer -> Integer
+fun2' = sum 
+ . filter even
+ . takeWhile (>1) 
+ . iterate (\n -> 
+    if even n
+      then n `div` 2
+      else 3 * n + 1)
```
shineli1984: we normally align if, then and else like this:
```Haskell
if even n
then n `div` 2
else 3 * n + 1
```

cht8687: @shineli1984 thanks!

Also uploaded another solution for this task. Plz check.

cis194/week4/robert-chang/src/Lib.hs
```diff
@@ -20,10 +20,35 @@ fun2 n | even n = n + fun2 (n `div` 2)
...
-      then n `div` 2
-      else 3 * n + 1) 
+    then n `div` 2
+    else 3 * n + 1) 
+-}
+
+fun2' = sum 
+ . filter even
+ . takeWhile (>1) 
+ . iterate (\n -> (even n ) ?: (n `div` 2, (3 * n + 1)))
```
shineli1984: If we can reduce the usage of brackets, that'll be better. Because brackets requires the reader to keep things in mind between them. And it gets worse when they are nested.

shineli1984:
one nit.

PR #58, opened 2019-09-02
Week 4 Ex 2, 3, 4

tienwei:
Took a while. Please review. Thanks!


cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -30,3 +30,34 @@ data Tree a
...
+insert x Leaf = Node 0 Leaf x Leaf
+insert x (Node i left c Leaf) = Node i left c (Node (i + 1) Leaf x Leaf)
+insert x (Node i Leaf c right) = Node i (Node (i + 1) Leaf x Leaf) c right
+insert x (Node i left c right)
+  | (checkDepth left) > (checkDepth right) = Node i left c (insert x right)
+  | otherwise = Node i (insert x left) c right
+
+foldTree :: (Foldable t, Eq a) => t a -> Tree a
+foldTree arr
+  | length arr == 0 = Leaf
```
shineli1984: Do we need this case here?

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -30,3 +30,34 @@ data Tree a
...
+
+-- Exercise 3.1 --
+xor :: [Bool] -> Bool
+xor = odd . foldr (\x acc -> acc + (bool 0 1 x)) (0 :: Integer)
+-- Exercise 3.2 --
+map' :: (a -> b) -> [a] -> [b]
+map' f = foldr (\x acc-> f x:acc) []
+-- Exercise 3.3 --
+myFoldl :: (a -> b -> a) -> a -> [b] -> a
+myFoldl f base xs = foldr (\x acc -> acc `f` x) base xs
```
shineli1984: is `foldl (flip (:)) [] [1, 2, 3]` the same as `myFoldl (flip (:)) [] [1, 2, 3]`?

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -30,3 +30,34 @@ data Tree a
...
+  | right /= Leaf && left == Leaf = 1 + checkDepth right
+  | otherwise = 1 + max (checkDepth left) (checkDepth right)
+
+insert :: Eq t => t -> Tree t -> Tree t
+insert x Leaf = Node 0 Leaf x Leaf
+insert x (Node i left c Leaf) = Node i left c (Node (i + 1) Leaf x Leaf)
+insert x (Node i Leaf c right) = Node i (Node (i + 1) Leaf x Leaf) c right
+insert x (Node i left c right)
+  | (checkDepth left) > (checkDepth right) = Node i left c (insert x right)
+  | otherwise = Node i (insert x left) c right
```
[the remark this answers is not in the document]
tienwei: My intension was to get the deepest level of left and right, and then insert the node to the one with fewer levels. Is this logic incorrect?

cis194/week4/tien/homework4/src/Homework4.hs
cht8687: @shineli1984 I cannot see any difference between the two.

shineli1984: @cht8687 because (-) is used in test for this function, we can not detect any difference because the arguments for (-) are both Num.

cis194/week4/tien/homework4/src/Homework4.hs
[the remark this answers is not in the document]
shineli1984: The tricky part I think is when the deepest level of left and right are the same, but there will be a difference when insert `t` to the right and to the left.
Like demostrated below, assume the 4th level are all leaves, the depth on the right and on the left are the same from the root node perspective. But in this case we can not insert a new node the right. 
```
N 
|
-----------
|         |
N         N
------    -------
|    |    |     |
N    L    N     N
.......

File: cis194/week4/tien/homework4/src/Homework4.hs [hunk 4]
```diff
@@ -30,3 +30,33 @@ data Tree a
   = Leaf
   | Node Integer (Tree a) a (Tree a)
   deriving (Show, Eq)
+
+checkDepth :: Eq a => Tree a -> Integer
+checkDepth Leaf = (0 :: Integer)
+checkDepth (Node _ Leaf _ Leaf) = 1
+checkDepth (Node _ left _ right)
+  | left /= Leaf && right == Leaf = 1 + checkDepth left
+  | right /= Leaf && left == Leaf = 1 + checkDepth right
```
shineli1984: Do we need `right /= Leaf` and `left /= Leaf` in these 2 guards?

cis194/week4/tien/homework4/src/Homework4.hs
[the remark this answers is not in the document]
tienwei: Hmm. You are right. I will try to fix this. Thanks a lot for the detailed explanation!

cis194/week4/tien/homework4/src/Homework4.hs
tienwei: Fixed.

cis194/week4/tien/homework4/src/Homework4.hs
tienwei: Refined the pattern matching.

cis194/week4/tien/homework4/src/Homework4.hs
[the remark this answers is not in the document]
tienwei: Hi @shineli1984, thanks for reviewing. I have updated the check. Now the tree should be balanced. Cheers.

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -30,3 +30,42 @@ data Tree a
...
+checkDepth (Node _ left@(Node _ _ _ _) _ Leaf) = 1 + checkDepth left
+checkDepth (Node _ left@(Node _ _ _ _) _ right@(Node _ _ _ _)) =
+  1 + checkDepth left + checkDepth right
+
+insert :: Eq t => t -> Tree t -> Tree t
+insert x Leaf = Node 0 Leaf x Leaf
+insert x (Node i left c Leaf) = Node i left c (Node (i + 1) Leaf x Leaf)
+insert x (Node i Leaf c right) = Node i (Node (i + 1) Leaf x Leaf) c right
+insert x (Node i left c right)
+  | (checkDepth left) > (checkDepth right) = Node i left c (insert x right)
```
shineli1984: This works. But here is something to think about. Is it better to still use the heights stored on the Node but treat left height equals to right height as a dedicated case?

cis194/week4/tien/homework4/src/Homework4.hs
```diff
@@ -30,3 +30,42 @@ data Tree a
...
+xor :: [Bool] -> Bool
+xor = odd . foldr (\x acc -> acc + (bool 0 1 x)) (0 :: Integer)
+
+-- Exercise 3.2 --
+map' :: (a -> b) -> [a] -> [b]
+map' f = foldr (\x acc -> f x : acc) []
+
+-- Exercise 3.3 --
+myFoldl :: (a -> b -> a) -> a -> [b] -> a
+myFoldl f base xs = foldr (\x acc -> f acc x) base (reverse xs)
```
shineli1984: Thought exercise here: reverse before folding is not efficient. Because it's 2 foldings on the same list. Is there a way to only fold it once? (hint: a recursive function is the same of a stack and stack operations (push and pop) are reversing the list of values when they are popped out comparing to when they are pushed in).  Is there a way to create a recursive function by folding out of a list then trigger it with `base`?

PR #59, opened 2019-09-03
Wei Ning - CIS194 Week4 Homework

powergun:
To run all the tests `stack test`. 

A note on testing:
Most exercises are tested in QuickCheck (Checks.hs), except for:
- exercise3 `xor` which is tested in Hspec (Specs.hs).
- exercise2 balanced binary tree folding, which is tested by printing out the height.

Thanks Shine for teaching!!


cis194/week4/weining/src/Exercise1.hs
```diff
@@ -0,0 +1,34 @@
...
+
+fun2Original :: Integer -> Integer
+fun2Original 1 = 0
+fun2Original n | even n    = n + fun2Original (n `div` 2)
+               | otherwise = fun2Original (3 * n + 1)
+
+fun2 n = sum $
+         filter even $
+         takeWhile (> 1) $
+         iterate (bool <$> ((+ 1) . (* 3)) <*> (flip div 2) <*> even) n
```
shineli1984: nit:
```Haskell
fun2 = sum
     . filter even
     . takeWhile (> 1)
     . iterate (bool <$> ((+ 1) . (* 3)) <*> (flip div 2) <*> even)
```
To turn function applications to function composition, the standard trick is to replace `$` with `.` then add `$` before the last variable. Because the last variable is the same as the only function input to fun2, we can make this point free like above.

cis194/week4/weining/src/Exercise1.hs
```diff
@@ -0,0 +1,34 @@
...
+
+fun1Original :: [Integer] -> Integer
+fun1Original [] = 1
+fun1Original (x:xs)
+    | even x = (x - 2) * fun1Original xs
+    | otherwise = fun1Original xs
+
+fun1 :: [Integer] -> Integer
+fun1 = (foldr (*) 1) .
+       (map (bool <$> (const 1) <*> (subtract 2) <*> even))
```
shineli1984: In Haskell we often align symbols. 
```Haskell
fun1 = foldr (*) 1 -- since function application has the highest precedence, we don't need brackets around this.
     . map calc
   where
     calc = bool <$> const 1 <*> subtract 2 <*> even -- we are trying to get rid of some brackets here to reduce noise in the code base
```

powergun: Nice style!

powergun:
Thank you soo much Shine!!!

PR #60, opened 2019-09-04
Exercise 3: More folds!

stevemao:


cis194/week4/stevemao/Homework.hs
```diff
@@ -0,0 +1,15 @@
...
+
+xor :: [Bool] -> Bool
+xor xs = 
+  foldr f False xs
+
+g :: (a -> b) -> a -> [b] -> [b]
+g f a list = [f a] ++ list
+
+map' :: (a -> b) -> [a] -> [b]
+map' f list = foldr (g f) [] list
```
shineli1984: Because the g above is not reused anywhere, there is no need for it to be a high order function. Instead, we can use a where clause instead.
```suggestion
map' f xs = foldr f' [] xs -- I used xs instead of list, this is a convention in Haskell
  where
    f' cur acc = f cur : acc -- I used cur, acc to indicate the current value and accumulated value.
```

cis194/week4/stevemao/Homework.hs
```diff
@@ -0,0 +1,15 @@
+module Homework where
+
+f :: Bool -> Bool -> Bool
+f False acc = acc
+f x y = x /= y
```
shineli1984: I think `f` is the same as `/=`

shineli1984:
Maybe we can try to use stack next time. The instruction is in the README.MD

PR #63, opened 2019-09-06
steve week 5

stevemao:


stevemao:
@shineli1984 I just added exercise 4

cis194/week5/stevemao/Var.hs
```diff
@@ -0,0 +1,35 @@
...
+import qualified Data.Map as M
+import Homework
+
+class HasVars a where
+  var :: String -> a
+
+data VarExprT = Lit Integer
+           | Add VarExprT VarExprT
+           | Mul VarExprT VarExprT
+           | Var Integer
```
shineli1984: To represent variable, it needs a name and we need to use `String` here at least to represent that.

cis194/week5/stevemao/Var.hs
```diff
@@ -0,0 +1,35 @@
...
+
+instance HasVars VarExprT where
+  var = Var
+
+instance HasVars (M.Map String Integer -> Maybe Integer) where
+  var = M.lookup
+
+instance Expr (M.Map String Integer -> Maybe Integer) where
+  lit a = (\_ -> Just a)
+  add a b = (\m -> fmap (+) (a m) <*> (b m))
```
shineli1984:
```suggestion
  add a b = (\m -> (+) <$> (a m) <*> (b m)) -- this is a common pattern when using applicative. <$> followed by one or more <*>
```

stevemao: When to use applicative and when to use `fmap` and `chain`?

shineli1984: When we can have `<$>` followed by one or `<*>`, we do `f <$> ... <*> ... <*> ...`.
`<$>` with `>>=` is not comparable to the above since they are with different type signature.
`f <$> a >>= g` need 2 functions `f` and `g`.

FYI, we call `>>=` `bind` in Haskell.

PR #65, opened 2019-09-06
Week2

hvariant:
no idea who the hacker was :/


cis194/week2/zhansongl/LogAnalysis.hs
```diff
@@ -0,0 +1,63 @@
...
+import qualified Text.Megaparsec.Char.Lexer as L
+import Data.Void
+
+type Parser = Parsec Void String
+
+messageTypeParser :: Parser MessageType
+messageTypeParser = do
+  charType <- oneOf "IEW"
+  case charType of
+       'I' -> return Info
```
shineli1984: NIT: Because GHC has now added Applicative as a constraint on Monad, we prefer `pure` instead of `return`.

cis194/week2/zhansongl/LogAnalysis.hs
```diff
@@ -0,0 +1,63 @@
...
+       'W' -> return Warning
+       'E' -> do 
+          space1
+          errorCode <- L.decimal
+          return (Error errorCode)
```
shineli1984: The do notation is equivalent to the following:
`space1 *> L.decimal <&> Error`
`*>` disregards the result of `space1` and `<&>` is a flipped version of `<$>`

cis194/week2/zhansongl/LogAnalysis.hs
```diff
@@ -0,0 +1,63 @@
...
+
+messageParser :: Parser LogMessage
+messageParser = do
+  messageType <- messageTypeParser
+  space1
+  timeStamp <- L.decimal
+  space
+  message <- many asciiChar
+  return $ LogMessage messageType timeStamp message
```
shineli1984: I think `messageParser` can be the below. 
```Haskell
LogMessage 
<$> (messageTypeParser <* space1) 
<*> (L.decimal <* space) 
<*> (many asciiChar)
```

hvariant:
Hmm, I'm finally beginning to understand how applicative operators work, lemme make a new PR to fix this

cis194/week2/zhansongl/LogAnalysis.hs
hvariant: yeah, saw this in the wiki after I finished writing this code, will fix

cis194/week2/zhansongl/LogAnalysis.hs
hvariant: hmm, it's more readable if you understand how it works, but less readable if you have no idea... will fix

cis194/week2/zhansongl/LogAnalysis.hs
hvariant: will fix

PR #66, opened 2019-09-07
Week3

hvariant:
no partial allowed


cis194/week3/zhansongl/Golf.hs
```diff
@@ -0,0 +1,58 @@
...
+- Create `h` number of rows where `h` is the maximum number of occurrences,
+- Draw `h` lines and put star or space in based on the occurrences of the digit
+-   and the line number, i.e. for the 2nd line if there are more than 2 2's in the
+-   input a star should be drawn, otherwise it should be a space.
+- Put the lines together to get the histogram.
+-}
+
+histogram :: [Int] -> String
+-- 182 chars
+histogram xs = (++ "==========\n0123456789\n") .
```
shineli1984: a nit on styling here. In Haskell, we normally align symbols on multiple lines:
```Haskell
histogram xs = (++ "==========\n0123456789\n")
             . unlines
             $ ...
```

shineli1984:
Good work!

cis194/week3/zhansongl/Golf.hs
hvariant: thanks, will fix

PR #67, opened 2019-09-07
Week 5 Ex 1, 2, 3, 4, 5

tienwei:
Not too sure about Ex3. Any feedback is greatly appreciated. 
Thanks.


cis194/week5/tien/src/Calc.hs
```diff
@@ -0,0 +1,74 @@
...
+
+import           ExprT
+import           Parser
+import           StackVM
+
+-- Exercise 1 --
+eval :: ExprT -> Integer
+eval (Lit a)         = a
+eval (ExprT.Add a b) = eval a + eval b
+eval (ExprT.Mul a b) = eval a * eval b
```
shineli1984: Instead of using `ExprT.Add`, we can import `ExprT` qualified. 
As in, `import qualified ExprT as ET` then, we are forced to use `ET.Mul`.
This is usually how we does it in production code base so that people know where the functions are coming from

cis194/week5/tien/src/Calc.hs
```diff
@@ -0,0 +1,74 @@
...
+
+-- Exercise 2 --
+evalStr :: String -> Maybe Integer
+evalStr (parseExp Lit ExprT.Add ExprT.Mul -> justExpr) = eval <$> justExpr
```
shineli1984: We are merely binding the result to a name `justExpr`. This means we are not using any view patterns features.
```suggestion
evalStr = fmap eval . parseExp Lit ExprT.Add ExprT.Mul
```

cis194/week5/tien/src/Calc.hs
```diff
@@ -0,0 +1,74 @@
...
+
+instance Expr Integer where
+  lit a = a
+  add a b = a + b
+  mul a b = a * b
+
+instance Expr Bool where
+  lit a = a > 0
+  add a b = a || b
+  mul a b = a && b
```
shineli1984: A pointfree version:
```Haskell
  lit = ( > 0 )
  add = ( || )
  mul = ( && )
```

cis194/week5/tien/src/Calc.hs
```diff
@@ -0,0 +1,74 @@
...
+
+instance Expr MinMax where
+  lit a = MinMax a
+  add (MinMax a) (MinMax b) = MinMax $ max a b
+  mul (MinMax a) (MinMax b) = MinMax $ min a b
+
+instance Expr Mod7 where
+  lit a = Mod7 $ mod a 7
+  add (Mod7 a) (Mod7 b) = Mod7 $ lit a + lit b
```
shineli1984: This line only works because we have defined `Integer` instance off `Expr a` typeclass.
The `+` used here says `lit a` and `lit b` must be of instance `Num a`. It just so happened that `Expr Integer` is defined earlier to make `lit a + lit b` possible because `Integer` is an instance of `Num a` type class.
But in order to fulfil the `modular` requirement, that is `(3 + 5) == 1`, we need make sure `a + b` is passed to lit instead.

cis194/week5/tien/src/Calc.hs
```diff
@@ -0,0 +1,83 @@
...
+  mul = (&&)
+
+instance Expr MinMax where
+  lit a = MinMax a
+  add (MinMax a) (MinMax b) = MinMax $ max a b
+  mul (MinMax a) (MinMax b) = MinMax $ min a b
+
+instance Expr Mod7 where
+  lit a = Mod7 $ mod a 7
+  add (Mod7 a) (Mod7 b) = Mod7 $ lit (a + b)
```
shineli1984: nit:
Normally, we use function composition style:
```suggestion
  add (Mod7 a) (Mod7 b) = Mod7 . lit $ (a + b)
```

shineli1984:
Good work!

PR #70, opened 2019-09-07
Week4

hvariant:


cis194/week4/zhansongl/Wholemeal.hs
```diff
@@ -0,0 +1,71 @@
...
+insert x Leaf = Node 0 Leaf x Leaf
+insert x (Node h left y right)
+  | getHeight left < getHeight right =
+      let h' = 1 + max (getHeight left') (getHeight right)
+          left' = insert x left
+      in  Node h' left' y right
+  | otherwise =
+      let h' = 1 + max (getHeight left) (getHeight right')
+          right' = insert x right
+      in  Node h' left y right'
```
shineli1984: For the case of `left` and `right` being the same height, is it correct to always insert on the `right`?

cis194/week4/zhansongl/Wholemeal.hs
```diff
@@ -0,0 +1,71 @@
...
+foldTree = foldr insert Leaf
+
+xor :: [Bool] -> Bool
+xor = odd . length . filter id
+
+map' :: (a -> b) -> [a] -> [b]
+map' f = foldr (\x ys -> (f x):ys) []
+
+myFoldl :: (a -> b -> a) -> a -> [b] -> a
+myFoldl f base = foldr (flip f) base . reverse
```
shineli1984: This is correct logic.
Now let's make it more efficient. Because `reverse` is doing a `folding` on the list. Plus `foldr`, we are folding twice here. Is there a way only folding once? (hint: how about folding out a recursive function then call with base?) The correct answer is the implementation of `foldl'`

cis194/week4/zhansongl/Wholemeal.hs
hvariant: it should be, because inserting a single element can increase the height of a tree by at most one, I think

shineli1984: When `left` and `right` being the same height, is it possible that inserting to the `left` does not increase the height and inserting to the `right` does?

hvariant: yes, in which case it will still be balanced because the height difference is 1

cis194/week4/zhansongl/Wholemeal.hs
hvariant: fixed, took me a while to figure this one out

cis194/week4/zhansongl/Wholemeal.hs
shineli1984: So following my previous comment, when it’s inserting to the right the new node underneath will be 2 levels deeper than the shorter leg under left. No?

hvariant: I think I understand what you mean now, will fix

hvariant: hmm, actually, I still think it's correct because if you look at the testing code [here](https://github.com/fp-works/2019-winter-Haskell-school/pull/70/files#diff-b43f2ea241398742a3d1444be175cc08R16) the resulting tree should be balanced given an arbitrary list which should be enough proof that it is correct.

hvariant: in the second case the height of the left tree should be equal or greater than the right, which means inserting to the right will increase the height difference by at most one, and will still result in a balanced tree

PR #71, opened 2019-09-08
Week5

hvariant:
not sure why we need to define VarExprT but...


cis194/week5/zhansongl/Calc.hs
```diff
@@ -0,0 +1,95 @@
...
+  lit = Mod7 . (`mod` 7)
+  add (Mod7 x) (Mod7 y) = Mod7 $ (x + y) `mod` 7
+  mul (Mod7 x) (Mod7 y) = Mod7 $ (x * y) `mod` 7
+
+testExp :: Expr a => String -> Maybe a
+testExp = parseExp lit add mul
+
+instance Expr S.Program where
+  lit n = [S.PushI n]
+  add p1 p2 = p1 ++ p2 ++ [S.Add]
```
stevemao: What if `p1` or `p2` is `[BVal True]`?

hvariant: hmm, I'm not sure actually, is this supposed to work with BVal?

stevemao: Oh I re-read the question again, looks like we should only worry about `IVal`...

The module `StackVM` is a bit misleading I think... Makes me overthink too much.

stevemao: > For any arithmetic expression exp :: Expr a => a it should be the case that
`stackVM exp == Right [IVal exp]`

I think this sentence is wrong `stackVM exp == Right [IVal exp]`

It should be `stackVM exp == Right [IVal val]` @shineli1984 Am I correct?

stevemao: Honestly, I can understand the question, but wasn't sure what I was supposed to do with this. I think this solution is correct though :)

cis194/week5/zhansongl/Calc.hs
```diff
@@ -0,0 +1,95 @@
...
+  add = VarAdd
+  mul = VarMul
+
+instance HasVars (M.Map String Integer -> Maybe Integer) where
+  var = M.lookup
+
+instance Expr (M.Map String Integer -> Maybe Integer) where
+  lit n = (\_ -> Just n)
+  add e1 e2 m = liftA2 (+) (e1 m) (e2 m)
+  mul e1 e2 m = liftA2 (*) (e1 m) (e2 m)
```
shineli1984: This comment is just to re-enforce your understanding of applicative:
```suggestion
  mul e1 e2 m = (*) <$> (e1 m) <*> (e2 m)
```

cis194/week5/zhansongl/Calc.hs
```diff
@@ -0,0 +1,95 @@
...
+
+instance Expr (M.Map String Integer -> Maybe Integer) where
+  lit = const . Just
+  add e1 e2 m = liftA2 (+) (e1 m) (e2 m)
+  mul e1 e2 m = liftA2 (*) (e1 m) (e2 m)
+
+withVars :: [(String, Integer)]
+         -> (M.Map String Integer -> Maybe Integer)
+         -> Maybe Integer
+withVars vs exp = exp $ M.fromList vs
```
shineli1984: function composition style is preferred:
```suggestion
withVars vs exp = exp . M.fromList $ vs
```

cis194/week5/zhansongl/Calc.hs
hvariant: I see, so <$> is the same as liftA

PR #76, opened 2019-09-08
Week 5 ex 2 to 6

zehua:
* added solutions to ex 2 to 6
* reverted the changes to `Parser.hs` and `StackVM.hs` in the parent folder and made copies of them (added `Eq` to `StackVal` and `StackExp`)

### DIFF
--- a/cis194/week5/zehua/src/Exercises.hs
+++ b/cis194/week5/zehua/src/Exercises.hs
@@ -1,9 +1,94 @@
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE FlexibleInstances #-}
 module Exercises where
 
-import           ExprT
+import qualified Data.Map   as M
+import           Control.Monad (liftM2)
+import qualified ExprT      as E
+import           Parser     (parseExp)
+import qualified StackVM    as S
 
 -- ex1
-eval :: ExprT -> Integer
-eval (Lit x)     = x
-eval (Add e1 e2) = eval e1 + eval e2
-eval (Mul e1 e2) = eval e1 * eval e2
+eval :: E.ExprT -> Integer
+eval (E.Lit x)     = x
+eval (E.Add e1 e2) = eval e1 + eval e2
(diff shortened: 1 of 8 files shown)


cis194/week5/zehua/src/Exercises.hs
```diff
@@ -1,9 +1,98 @@
...
+instance HasVars (M.Map String Integer -> Maybe Integer) where
+  var = M.lookup
+
+instance Expr (M.Map String Integer -> Maybe Integer) where
+  lit = const . Just
+  -- add f1 f2 m = liftM2 (+) (f1 m) (f2 m)
+  -- add f1 f2 = liftM2 (+) <$> f1 <*> f2
+  -- add f1 f2 = liftA2 (liftM2 (+)) f1 f2
+  add = liftA2 . liftM2 $ (+)
+  mul = liftA2 . liftM2 $ (*)
```
zehua: Is there a way to do something similar to:
```
    add = f (+)
    mul = f (*)
  where
    f = liftA2 . liftM2
```

cis194/week5/zehua/src/Exercises.hs
```diff
@@ -1,9 +1,99 @@
...
+  var = M.lookup
+
+instance Expr (M.Map String Integer -> Maybe Integer) where
+  lit = const . Just
+  -- map (+) over Maybe
+  -- add f1 f2 m = liftA2 (+) (f1 m) (f2 m)
+  -- map (liftA2 (+)) over reader applicative
+  -- add f1 f2 = liftA2 (+) <$> f1 <*> f2
+  -- add f1 f2 = liftA2 (liftA2 (+)) f1 f2
+  add = liftA2 . liftA2 $ (+)
```
shineli1984: Good one here.
FYI. It’s also acceptable/common that we write in the fmap ap ap ... style. As in 
`add  f1 f2 m = (+) <$> (f1 m) <*> (f2 m)`

PR #77, opened 2019-09-09
Week5-T1,T2,T3,T4

cht8687:


cis194/week5/robert-chang/src/Calc.hs
```diff
@@ -0,0 +1,47 @@
...
+newtype MinMax = MinMax Integer deriving (Eq,Show)
+instance Expr MinMax where
+  lit = MinMax
+  add (MinMax x) (MinMax y) = MinMax (max x y)
+  mul (MinMax x) (MinMax y) = MinMax (min x y)
+
+newtype Mod7 = Mod7 Integer deriving (Eq, Show)
+instance Expr Mod7 where
+  lit x = Mod7 (x `mod` 7)
+  add (Mod7 x) (Mod7 y) = Mod7 ((x + y) `mod` 7)
```
shineli1984: we have the `mod 7` logic on the lit, can we reuse it here?

PR #79, opened 2019-09-10
Daniel Deng Week04 - Exercise 02 ~ 04

sinogermany:


cis194/week4/daniel-deng/src/Exercise04.hs
```diff
@@ -0,0 +1,16 @@
+module CIS194.Homework04.Exercise04 (sieveSundaram) where
+
+cartProd :: [a] -> [(a, a)]
+cartProd xs = [(a, b) | a <- xs, b <- xs]
+
+validNumber :: Integer -> Bool
+validNumber x = not
```
sinogermany: @shineli1984 any idea how we can implement this in a point-free style here? Does point-free style make sense here from a pragmatic perspective?

shineli1984: I think this is good enough to convey the intention.

PR #82, opened 2019-09-12
Fibonacci

stevemao:


cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+fib 1 = 1
+fib n = (fib $ n - 1) + (fib $ n - 2)
+
+fibs1 :: [Integer]
+fibs1 = fmap fib [0..]
+
+fib' :: Integer -> Integer
+fib' 0 = 0
+fib' 1 = 1
+fib' n = snd $ foldr (\_ acc -> (snd acc, fst acc + snd acc)) (0, 1) [2..n]
```
shineli1984: Question: in order to generating `0, 1, 1, 2, 3` that is the first 5 numbers. How many times is addition invoked?

cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+fibs2 :: [Integer]
+fibs2 = fmap fib' [0..]
+
+data Stream a = Cons a (Stream a)
+
+streamToList :: Stream a -> [a]
+streamToList (Cons a s) = (a : streamToList s)
+
+instance Show a => Show (Stream a) where
+  show s = show . take 20 $ streamToList s
```
shineli1984:
```suggestion
  show = show . take 20 . streamToList
```

cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+
+streamFromList :: [a] -> Stream a
+streamFromList (x: xs) = Cons x $ streamFromList xs
+streamFromList _ = error "List must not be empty"
```
shineli1984: We must avoid partial function at all cost.

cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+
+streamRepeat :: a -> Stream a
+streamRepeat a = streamFromList $ repeat a
```
shineli1984:
```suggestion
streamRepeat = streamFromList . repeat
```

cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+
+streamMap :: (a -> b) -> Stream a -> Stream b
+streamMap f s = streamFromList . fmap f . streamToList $ s
```
shineli1984: Please do a pattern match on `s`. We don't necessary need to convert stream to a list here.

cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+
+streamFromSeed :: (a -> a) -> a -> Stream a
+streamFromSeed f = streamFromList . iterate f
+
+nats :: Stream Integer
+nats = streamFromList [0..]
```
shineli1984: Because `steamFromList` needs to be changed to have a type that does not fit here, we can utilise `streamFromSeed` above.

cis194/week6/stevemao/Fibonacci.hs
```diff
@@ -0,0 +1,55 @@
...
+
+merge :: [a] -> [a] -> [a]
+merge [] ys = ys
+merge (x:xs) ys = x:merge ys xs
+
+interleaveStreams :: Stream a -> Stream a -> Stream a
+interleaveStreams a b = streamFromList . merge (streamToList a) $ (streamToList b)
```
shineli1984: Same suggestion here that we don't necessarily need to convert a stream to list to merge it.

cis194/week6/stevemao/Fibonacci.hs
stevemao: twice?

shineli1984: I think it's going to be more than 3 times.

stevemao: I still don't see duplicated additions. It should just do the following ones:
```haskell
0 + 1
1 + 1
2 + 1
```

shineli1984: To get to the fibonacci number at position n, yes. 
How about using `fib'` to generate the fibonacci numbers from 0 to the number at position n.

shineli1984:
good work!

PR #85, opened 2019-09-14
week 6

zehua:


cis194/week6/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,115 @@
...
+x = Cons 0 $ Cons 1 $ streamRepeat 0
+
+instance Num (Stream Integer) where
+  fromInteger n                 = Cons n (streamRepeat 0)
+  negate (Cons n xs)            = Cons (-n) xs
+  (Cons a0 a') + (Cons b0 b')   = Cons (a0 + b0) $ a' + b'
+  (Cons a0 a') * b@(Cons b0 b') = Cons (a0 * b0) $ (fromInteger a0) * b' + a' * b
+
+--  (a0/b0) + x((1/b0)(A' − Q*B')).
+-- divStream :: Integral a => Stream a -> Stream a -> Stream a
```
zehua: This one does not work. Giving the following error:

```
    • Couldn't match expected type ‘Integer’ with actual type ‘a’
      ‘a’ is a rigid type variable bound by
        the type signature for:
          divStream :: forall a.
                       Integral a =>
                       Stream a -> Stream a -> Stream a
        at src/Exercises.hs:70:1-59
    • In the first argument of ‘fromInteger’, namely ‘b0’
      In the second argument of ‘divStream’, namely ‘(fromInteger b0)’
      In the first argument of ‘(*)’, namely
        ‘(divStream (fromInteger 1) (fromInteger b0))’
    • Relevant bindings include
        q :: Stream a (bound at src/Exercises.hs:74:7)
        b' :: Stream a (bound at src/Exercises.hs:72:33)
        b0 :: a (bound at src/Exercises.hs:72:30)
        a' :: Stream a (bound at src/Exercises.hs:72:20)
        a0 :: a (bound at src/Exercises.hs:72:17)
        divStream :: Stream a -> Stream a -> Stream a
          (bound at src/Exercises.hs:72:1)
   |
75 |           (divStream (fromInteger 1) (fromInteger b0)) * (a' - q * b')
   |                                                   ^^
```

Looks like `fromInteger a` could not treat `Integral` as `Integer`

shineli1984: `Integral` typeclass requires `Enum` and `Real` as constraints which `Stream a` does not have instance for.

zehua: But the error msg complains about `a` instead of `Stream a`. `b0` is Integer` which is an instance of `Integral`.

zehua: ok. I think the error message from ghc says exactly what is wrong. `fromInteger` expects exactly the data type `Integer` but we are passing in an `Integral a => a`, which `Integer` is an instance of but they are not the same thing.

zehua: created #87 to make this work

cis194/week6/zehua/src/Exercises.hs
```diff
@@ -0,0 +1,115 @@
...
+        (a11 * b12 + a12 * b22)
+        (a21 * b11 + a22 * b21)
+        (a21 * b12 + a22 * b22)
+
+f1 :: Matrix Integer
+f1 = Matrix 1 1 1 0
+
+fib4 :: Integer -> Integer
+-- taking power of (n+1) to handle fib4 0 == 0
+fib4 n = case f1 ^ (n + 1) of
```
stevemao: Why not `case f1 ^ n` and `Matrix _ a _ _ -> a`?

zehua: that does not work with `fib4 0`. `f1 ^ 0` seems to be resolved as `f1`.

stevemao: Is it common to scarifies performance to write one less pattern matching in Haskell? @shineli1984

sinogermany:
```haskell
fib4 :: Integer -> Integer
fib4 n
  | n == 0 .  = 0 -- this is the definition of fibs
  | otherwise = ...
```

stevemao: Yeah that's what I was asking. If you do one more addition, you don't have to declare `n == 0    = 0`. I'm not sure which way is better.

PR #89, opened 2019-09-16
add wk4 Exercise 1: Wholemeal programming

stevemao:


cis194/week4/stevemao/Homework.hs
```diff
@@ -1,6 +1,33 @@
 {-# OPTIONS_GHC -Wall #-}
 module Homework where
 
+fun1 :: [Integer] -> Integer
+fun1 [] = 1
+fun1 (x:xs)
```
sinogermany: Can we get rid of the recursion calls here?

sinogermany: - `otherwise = fun1 xs` means `if x is not even, I'll skip all the way to the next even element and apply fun1 on it`
- `even x = (x - 2) * fun1 xs` means `if x is even, I'll subtract x with 2 and multiply it by the next value`
So `func1 === "subtract all the even numbers by 2 and multiply the results together"`

stevemao: This is in the original Question I just copied it here. My solution is below: `fun1'`.

sinogermany: > This is in the original Question I just copied it here. My solution is below: `fun1'`.

Ops.. my bad. Sorry...

cis194/week4/stevemao/Homework.hs
```diff
@@ -1,6 +1,33 @@
...
+fun2 n | even n    = n + fun2 (n `div` 2)
+       | otherwise = fun2 (3 * n + 1)
+
+fn2 :: Integer -> Integer
+fn2 a 
+  | even a = a `div` 2
+  | otherwise = 3 * a + 1
+
+fun2' :: Integer -> Integer
+fun2' n = sum . filter even . takeWhile (> 1) . iterate fn2 $ n
```
shineli1984:
```suggestion
fun2' = sum . filter even . takeWhile (> 1) . iterate fn2
```

PR #91, opened 2019-09-18
Daniel M week6 homework

danielmcmillan:


cis194/week6/daniel-mc/Fibonacci.hs
```diff
@@ -0,0 +1,103 @@
...
+
+streamToList :: Stream a -> [a]
+streamToList (Stream x xs) = x : streamToList xs
+
+instance Show a => Show (Stream a) where
+  show = show . take 20 . streamToList
+
+-- Exercise 4
+streamRepeat :: a -> Stream a
+streamRepeat x = Stream x (streamRepeat x)
```
sinogermany: FYI - if you would like to have a point-free version here:

```suggestion
streamRepeat = Stream <*> streamRepeat
```

cis194/week6/daniel-mc/Fibonacci.hs
```diff
@@ -0,0 +1,103 @@
...
+
+streamMap :: (a -> b) -> Stream a -> Stream b
+streamMap f (Stream x xs) = Stream (f x) (streamMap f xs)
+
+streamFromSeed :: (a -> a) -> a -> Stream a
+streamFromSeed f x = Stream x (streamFromSeed f (f x))
```
sinogermany:
```suggestion
streamFromSeed f x = Stream x . streamFromSeed f $ f x
```

(I think it'll work)

cis194/week6/daniel-mc/Fibonacci.hs
```diff
@@ -0,0 +1,103 @@
...
+
+instance Show a => Show (Stream a) where
+  show = show . take 20 . streamToList
+
+-- Exercise 4
+streamRepeat :: a -> Stream a
+streamRepeat x = Stream x (streamRepeat x)
+
+streamMap :: (a -> b) -> Stream a -> Stream b
+streamMap f (Stream x xs) = Stream (f x) (streamMap f xs)
```
sinogermany: Not sure which one's better. Is it a subjective thing? @shineli1984 

```suggestion
streamMap f (Stream x xs) = Stream (f x) $ streamMap f xs
```

cis194/week6/daniel-mc/Fibonacci.hs
stevemao: `streamFromSeed f x = Stream x . streamFromSeed f . f $ x`

cis194/week6/daniel-mc/Fibonacci.hs
danielmcmillan: Not sure but I think we prefer $

cis194/week6/daniel-mc/Fibonacci.hs
danielmcmillan: Cool thanks 😄

cis194/week6/daniel-mc/Fibonacci.hs
danielmcmillan: Thanks, but I might leave it out for now since I know very little about things like Applicative and Functor.

cis194/week6/daniel-mc/Fibonacci.hs
shineli1984: The general rule is that we want things gluing functions and/or variables together to be as succinct/invisible as possible. So that we can express ourselves by building bigger pieces from tiny ones (or even almost nothing) without introduce too much visual distraction. That's why `.` is a perfect fit as a composition operator. 
Here, `$` is preferred following the same rule, because `(`,`)` are more visible than a single character `$`.

PR #92, opened 2019-09-18
Week6: use more applicative style

hvariant:


cis194/week6/zhansongl/Fibonacci.hs
```diff
@@ -14,7 +16,7 @@ fibs1 = map fib [0..]
 
 fibs2 :: [Integer]
 fibs2 = next 0 1
-  where next a b = a : next b (a+b)
+  where next n = (n:) . (<*>) next (n+)
```
shineli1984: Composing a infix operator such as `<*>`, `<$>`, `>>=` probably is a sign we are turning the nob too high on composition. But it's a fun exercise.

cis194/week6/zhansongl/Fibonacci.hs
```diff
@@ -25,13 +27,13 @@ instance Show a => Show (Stream a) where
   show = show . take 20 . streamToList
 
 streamRepeat :: a -> Stream a
-streamRepeat a = Cons a . streamRepeat $ a
+streamRepeat = Cons <*> streamRepeat
```
shineli1984: This is a perfect candidate for `<*>`

cis194/week6/zhansongl/Fibonacci.hs
```diff
@@ -25,13 +27,13 @@ instance Show a => Show (Stream a) where
...
 
 streamMap :: (a -> b) -> Stream a -> Stream b
 streamMap f (Cons h t) = Cons (f h) (streamMap f t)
 
 streamFromSeed :: (a -> a) -> a -> Stream a
-streamFromSeed f s = Cons s (streamFromSeed f (f s))
+streamFromSeed f = Cons <*> (streamFromSeed f . f)
```
shineli1984:
```suggestion
streamFromSeed f = Cons <*> streamFromSeed f . f
```
A lot operators are designed to have the precedence just right, so that we don't have to put brackets around function compositions before and after. `<*>` is one of them.

cis194/week6/zhansongl/Fibonacci.hs
hvariant: I was trying to figure out if there is a way to remove n as well but was not able to... but yeah, this level of applicative style is probably a bit excessive

shineli1984: pointfree.io？

hvariant: oh wow

PR #97, opened 2019-09-20
W7steve

stevemao:


stevemao:
Not sure if this is correct but it compiles

cht8687:
where is your tests?

cis194/week7/stevemao/Scrabble.hs
```diff
@@ -0,0 +1,68 @@
...
+  | toUpper a `elem` ['D', 'G'] = Score 2
+  | toUpper a `elem` ['B', 'C', 'M', 'P'] = Score 3
+  | toUpper a `elem` ['F', 'H', 'V', 'W', 'Y'] = Score 4
+  | toUpper a == 'K' = Score 5
+  | toUpper a `elem` ['J', 'X'] = Score 8
+  | toUpper a `elem` ['Q', 'Z'] = Score 10
+score _ = Score 0
+
+scoreString :: String -> Score
+scoreString = sum . fmap score
```
shineli1984: Thought exercise:
Is it the same to write `scoreString = foldMap score`? and why

cis194/week7/stevemao/JoinList.hs
```diff
@@ -0,0 +1,53 @@
...
+  | i < getIntSize s = indexJ i a
+  | otherwise = indexJ i b
+
+dropJ :: (Sized b, Monoid b) =>
+  Int -> JoinList b a -> JoinList b a
+dropJ _ Empty = Empty
+dropJ i l@(Single s _)
+  | i < getIntSize s = l
+  | otherwise = Empty
+dropJ i (Append s a b)
```
shineli1984: I'm trying to understand the logic here. 
Does `s` stand for the sum of the sizes for `a` and `b`? if so, does `i < getIntSize s` mean we are dropping anything up to the total size? and `i > getIntSize` means we are over-dropping?

cis194/week7/stevemao/Scrabble.hs
stevemao: Yeah, it maps each element of the structure to a monoid, and combine (`mappend`) the results.

cis194/week7/stevemao/JoinList.hs
stevemao: No, it may or may not over-dropping but I know that the first dropped element will be in `b`. But `b` is another 0 based index so I need to do `i - getSize a`.

stevemao: last dropped element*

cis194/week7/stevemao/JoinList.hs
```diff
@@ -20,34 +20,39 @@ tag (Append m _ _) = m
...
-  | i < getIntSize s = l
+dropJ i l@(Single _ _)
+  | i < getIntSize l = l
   | otherwise = Empty
-dropJ i (Append s a b)
-  | i < getIntSize s = dropJ i a +++ b
-  | i > getIntSize s = dropJ (i - getSize a) b
+dropJ i (Append _ a b)
+  | i < getIntSize a = dropJ i a +++ b
+  | i > getIntSize a = dropJ i b
```
shineli1984: Is it correct to drop i number of elements from b here?

cis194/week7/stevemao/JoinList.hs
```diff
@@ -20,34 +20,39 @@ tag (Append m _ _) = m
...
-  | i >= getIntSize s = l
+takeJ i l@(Single _ _)
+  | i >= getIntSize l = l
   | otherwise = Empty
-takeJ i (Append s a b)
-  | i < getIntSize s = takeJ i a
-  | i > getIntSize s = a +++ takeJ i b
+takeJ i (Append _ a b)
+  | i < getIntSize a = takeJ i a
+  | i > getIntSize a = a +++ takeJ i b
```
shineli1984: Same here

PR #98, opened 2019-09-21
fix the provided exercise files in week 7

zehua:
fixing a few issues that were only introduced in recent version of GHC.

The only warning remaining:

```
cis194/week7/zehua/src/StringBuffer.hs:(8,1)-(16,31): warning: [-Worphans]
    Orphan instance: instance Buffer String
    To avoid this
        move the instance declaration to the module of the class or of the type, or
        wrap the type with a newtype and declare the instance on the new type.
  |
8 | instance Buffer String where
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```


cis194/week7/Sized.hs
```diff
@@ -23,6 +24,8 @@ instance Sized Size where
 instance Sized b => Sized (a,b) where
   size = size . snd
 
+instance Semigroup Size where
```
zehua: > NOTE: Semigroup is a superclass of Monoid since base-4.11.0.0.

```
week7/zehua/src/Sized.hs:26:10-20: error:
    • No instance for (Semigroup Size)
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘Monoid Size’
   |
26 | instance Monoid Size where
   |          ^^^^^^^^^^^
```

cis194/week7/Editor.hs
```diff
@@ -37,6 +38,8 @@ commands = map show [View, Edit, Next, Prev, Quit]
 newtype Editor b a = Editor (StateT (b,Int) IO a)
   deriving (Functor, Monad, MonadIO, MonadState (b,Int))
 
+instance Applicative (Editor b) where
```
zehua:
```
cis194/week7/zehua/src/Editor.hs:38:22: error:
    • No instance for (Applicative (Editor b))
        arising from the 'deriving' clause of a data type declaration
      Possible fix:
        use a standalone 'deriving instance' declaration,
          so you can specify the instance context yourself
    • When deriving the instance for (Monad (Editor b))
   |
38 |   deriving (Functor, Monad, MonadIO, MonadState (b,Int))
   |                      ^^^^^
```

cis194/week7/Sized.hs
```diff
@@ -1,7 +1,8 @@
 {-# LANGUAGE GeneralizedNewtypeDeriving, FlexibleInstances #-}
+{-# OPTIONS_GHC -fno-warn-missing-methods #-}
 module Sized where
 
-import Data.Monoid
+-- import Data.Monoid
```
zehua:
```
week7/zehua/src/StringBuffer.hs:4:1: warning: [-Wunused-imports]
    The import of ‘Data.Monoid’ is redundant
      except perhaps to import instances from ‘Data.Monoid’
    To import instances alone, use: import Data.Monoid()
  |
4 | import Data.Monoid
  | ^^^^^^^^^^^^^^^^^^
```

stevemao:
Wow thanks!

zehua:
good suggestion for `(<>) = (+)` for `Semigroup Size`, @stevemao   👍

zehua:
The only warning remaining:

```
cis194/week7/zehua/src/StringBuffer.hs:(8,1)-(16,31): warning: [-Worphans]
    Orphan instance: instance Buffer String
    To avoid this
        move the instance declaration to the module of the class or of the type, or
        wrap the type with a newtype and declare the instance on the new type.
  |
8 | instance Buffer String where
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

zehua:
Not really sure if it's worthwhile to do 37856f6 just to remove the orphan instance warning.

shineli1984:
You can leave the orphan instance there and suppress the warning.

PR #99, opened 2019-09-21
Week7

hvariant:


cis194/week7/zhansongl/Scrabble.hs
```diff
@@ -0,0 +1,56 @@
...
+                   , ('T', 1)
+                   , ('W', 4)
+                   , ('V', 4)
+                   , ('Y', 4)
+                   , ('X', 8)
+                   , ('Z', 10)
+                   ]
+
+scoreString :: String -> Score
+scoreString = Score . getSum . foldMap (Sum . getScore . score)
```
shineli1984: `Score` is already a `Monoid` with `+` and `Score 0`. Can we leverage that directly instead of using `Sum` as an indirection here?

hvariant: right, I was writing `foldMap . score` but keep getting compiler errors, turns out it should be `foldMap score`...

anyways the fix will be in a new PR.

PR #104, opened 2019-09-24
Function Composition Cheatsheet

sinogermany:
@shineli1984 and everyone, keen to get input from you guys. Cheers. 

@shineli1984 thanks for all the help during the process. Much appreciated!


doc/function-composition-cheatsheet.md
```diff
@@ -0,0 +1,172 @@
...
+f = g <$> f1 <*> f2 <*> f3 <*> fn
+f = g . f1 <*> f2 <*> f3 <*> fn
+--
+```
+
+```haskell
+f x = g (f1 x) (f2 x) (f3 x) (fn x) x
+
+-- Applicative Style
+f = g <$> f1 <*> f2 <*> f3 <*> fn <*> id
```
sinogermany: @shineli1984 is it worth the effort of flipping for the purpose of removing the `id` here?

cht8687:
This is fantastic. keep up the great work.

doc/function-composition-cheatsheet.md
```diff
@@ -0,0 +1,172 @@
...
+f = f1 <<< f2 <<< f3 <<< fn
+f = fn >>> f3 >>> f2 >>> f1
+
+-- Flow
+f = f1 <. f2 <. f3 <. fn
+f = fn .> f3 .> f2 .> f1
+```
+
+```haskell
+-- Functor Law
```
shineli1984: Can we use `fmap` or `<$>` to state the functor law on function?

doc/function-composition-cheatsheet.md
shineli1984: Yes. A flipped g is usually what we expect when using <*>. Otherwise, we would use =<<

doc/function-composition-cheatsheet.md
```diff
@@ -0,0 +1,172 @@
...
+f a = f2 =<< f1 a
+f = f2 <=< f1
+```
+
+```haskell
+findUser :: UserID -> Either Err User
+getDepartment :: User -> Either Err Department
+getManager :: Department -> Either Err User
+
+getManagerByUserID -> UserID -> Either Err User
```
shineli1984:
```suggestion
getManagerByUserID :: UserID -> Either Err User
```

doc/function-composition-cheatsheet.md
sinogermany: Thanks @shineli1984 . 

Question though, if we have:
```haskell
f :: a -> b -> c -> d -> e
```

Then we would have:
```haskell
flip f :: b -> a -> c -> d -> e
```
So how do we solve this without using `id`?
```haskell
f x = g (f1 x) (f2 x) (f3 x) ... (fn x) x
```

shineli1984: Sorry. I wasn’t clear. I meant we would expect g to have x as the first argument

stevemao:
Should we create another repo for this?

PR #106, opened 2019-09-26
W7steve

stevemao:


stevemao:
@cht8687 boss, I'm adding tests now

cht8687:
@stevemao how did you finish the tasks without a test? 😄 it's a miracle.

stevemao:
I tried not to rely on test but focus on the types.

cht8687:
🤔

PR #108, opened 2019-09-27
Daniel Deng Week01 Rev01

sinogermany:
Let's see how much we've learned over time by looking back :) 
@shineli1984 Please let us know if there are better ways to address those (or we should've used pattern matching in some case). Thanks.


cis194/week1/daniel-deng/src/CIS194/Homework01/Exercise02.hs
```diff
@@ -1,9 +1,8 @@
...
-doubleEveryOtherFromLeft []                      = []
-doubleEveryOtherFromLeft [element]               = [element]
-doubleEveryOtherFromLeft (first : second : rest) = [first, second * 2] ++ doubleEveryOtherFromLeft rest
+import Data.Bool
 
 doubleEveryOther :: [Integer] -> [Integer]
-doubleEveryOther = reverse . doubleEveryOtherFromLeft . reverse
+doubleEveryOther = foldr f []
+                 where
+                   f x = flip (:) <*> bool x (x *2) . odd . length
```
sinogermany: @shineli1984 which one do you reckon is better? Is that subjective or there's a rule as a guidance?

```suggestion
                   f x = (:) =<< bool x (x *2) . odd . length
```

hvariant: This actually turns the function quadratic because `length` has to go through the whole list.

So... I like the original version better

sinogermany: > This actually turns the function quadratic because `length` has to go through the whole list.
> 
> So... I like the original version better

That depends on how Haskell runtime works on list length, worst case is O(n).
How about this anyway:

```haskell
doubleEveryOther :: [Integer] -> [Integer]
doubleEveryOther = snd . foldr f (0, [])
                 where
                   f x (c, xs) = (c + 1, newVal x c : xs)
                   newVal x    = bool x (x * 2) . odd
```

shineli1984:
These are simple enough that recursion does not bring much complexity.

sinogermany:
> These are simple enough that recursion does not bring much complexity.

Do you mean logical complexity or time complexity? Recursions seem to be logically less complexed to me lol.

PR #110, opened 2019-09-27
w8

stevemao:


cis194/week8/stevemao/Party.hs
```diff
@@ -0,0 +1,49 @@
...
+moreFun gl1 gl2
+  | compare gl1 gl2 == GT = gl1
+  | otherwise = gl2
+
+treeFold :: (a -> [b] -> b) -> Tree a -> b
+treeFold f (Node a forest) = f a . fmap (treeFold f) $ forest
+
+nextLevel :: Employee -> [(GuestList, GuestList)]
+  -> (GuestList, GuestList)
+nextLevel boss xs = foldr (\(gl1, gl2) (gl1', gl2') -> (glCons boss gl1 <> gl1', gl2 <> gl2')) (mempty, mempty) xs
```
stevemao: This logic is wrong... will fix it later.

cis194/week8/stevemao/Party.hs
```diff
@@ -0,0 +1,49 @@
...
+instance Semigroup GuestList where
+  GL el1 f1 <> GL el2 f2 = GL (el1 <> el2) (f1 + f2)
+
+instance Monoid GuestList where
+  mempty = GL [] 0
+
+moreFun :: GuestList -> GuestList -> GuestList
+moreFun gl1 gl2
+  | compare gl1 gl2 == GT = gl1
+  | otherwise = gl2
```
shineli1984: We need something like `Ord a => a -> a -> a`. Do we have a function that we can use under `Ord` typeclass?

cis194/week8/stevemao/Party.hs
```diff
@@ -0,0 +1,49 @@
...
+
+treeFold :: (a -> [b] -> b) -> Tree a -> b
+treeFold f (Node a forest) = f a . fmap (treeFold f) $ forest
+
+nextLevel :: Employee -> [(GuestList, GuestList)]
+  -> (GuestList, GuestList)
```
shineli1984: We usually align type like this:
```Haskell
nextLevel :: Employee 
          -> [ ( GuestList, GuestList ) ]
          -> ( GuestList, GuestList )
```

cis194/week8/stevemao/Party.hs
```diff
@@ -0,0 +1,49 @@
...
+nextLevel boss xs = foldr (\(gl1, gl2) (gl1', gl2') -> (glCons boss gl1 <> gl2', gl2 <> gl1')) (mempty, mempty) xs
+
+moreFunTuple :: (GuestList, GuestList) -> GuestList
+moreFunTuple = uncurry moreFun
```
shineli1984: If we only use uncurried `moreFunTuple` once, we don't need to bind it to a name. If we have to, we can just use `moreFunTuple'`

cis194/week8/stevemao/Party.hs
```diff
@@ -0,0 +1,49 @@
...
+moreFun gl1 gl2
+  | compare gl1 gl2 == GT = gl1
+  | otherwise = gl2
+
+treeFold :: (a -> [b] -> b) -> Tree a -> b
+treeFold f (Node a forest) = f a . fmap (treeFold f) $ forest
+
+nextLevel :: Employee -> [(GuestList, GuestList)]
+  -> (GuestList, GuestList)
+nextLevel boss xs = foldr (\(gl1, gl2) (gl1', gl2') -> (glCons boss gl1 <> gl2', gl2 <> gl1')) (mempty, mempty) xs
```
shineli1984: Have we tested this logic?

PR #111, opened 2019-09-29
Week7: Task[1..2]

cht8687:


cis194/week7/haotian/week7/src/JoinList.hs
```diff
@@ -0,0 +1,55 @@
...
+tag :: Monoid m => JoinList m a -> m
+tag Empty = mempty
+tag (Single m _) = m
+tag (Append m _ _) = m
+
+-- Readings:
+-- (<>) :: Monoid m => m -> m -> m
+-- (<>) = mappend
+(+++) :: Monoid m => JoinList m a -> JoinList m a -> JoinList m a
+(+++) l1 l2 = Append (tag l1 <> tag l2) l1 l2
```
stevemao: What about the id law? `mempty <> jl == jl == jl <> mempty`?

PR #114, opened 2019-09-30
Daniel Deng Week 07 - Exercise 04 WIP - Do Not Review Yet

sinogermany:
@shineli1984 ready to be reviewed. 
The meaty commits are: 

- https://github.com/fp-works/2019-winter-Haskell-school/commit/0f763f1bb53c6b2f7b744730f04b4bf264d0957d
- https://github.com/fp-works/2019-winter-Haskell-school/pull/114/commits/17884841e56a276ab6d8077c65b97a1ba031775d

Thanks :)


cis194/week7/daniel-deng/src/CIS194/Homework07/JoinList.hs
```diff
@@ -0,0 +1,26 @@
+module CIS194.Homework07.JoinList ( JoinList(..) ) where
+
+import CIS194.Homework07.Sized ( Sized(..) )
+
+data JoinList m a = Empty
+                  | Single m a
+                  | Append m (JoinList m a) (JoinList m a)
+  deriving (Eq, Show)
+
+tag :: Monoid m => JoinList m a -> m
```
sinogermany: @shineli1984 any way we could derive the monoidal `m` for `JoinList`? For `m` is the only thing in the type constructor that has a `Monoid` constraint.

cis194/week7/daniel-deng/src/CIS194/Homework07/JoinList.hs
```diff
@@ -0,0 +1,26 @@
...
+tag Empty          = mempty
+tag (Single m _)   = m
+tag (Append m _ _) = m
+
+instance Monoid m => Semigroup (JoinList m a) where
```
sinogermany: @shineli1984 I'm a bit confused here. 

```haskell
  Empty <> jl = jl
  jl <> Empty = jl
```

Shouldn't the definition of `mempty` take care of it when defining the `Monoid`?

I mean I understand that :

- `Semigroup` can be existing without `mempty`
- we need exhaustive pattern matching when defining `Semigroup` 
- we need that definition to declare the `Monad` instance

But does that mean basically any `Monoid` instances that needs pattern matching for `<>` needs to do left and right identity associativity? 

To me it's like saying:
```haskell
instance Semigroup A where
  (<>) :: A -> A -> A
  Empty <> x = x  -- Appending Empty with x would result in x itself
  x <> Empty = x  -- Appending x with Empty would result in x itself
  x1 <> x2 = blah -- Appending y1 and y2 would be whatever it would be

instance Monoid A where
  mempty :: A
  mempty = Empty -- BTW, deal readers, `Empty` is the identity element...
  
  -- So can haskell please take care of the left and right associativity for me?
```

That looks a bit verbose to me. Am I missing any language features? Or is that something that is just expected to be that verbose?

cis194/week7/daniel-deng/src/CIS194/Homework07/JoinList.hs
shineli1984: It seems we should be able to derive this. There must be some generic monoid module already doing this, but it's not in the GHC.

cis194/week7/daniel-deng/src/CIS194/Homework07/JoinList.hs
shineli1984: Like you said, by `take care of the left and right ... for me`, GHC would have to produce non-exhaustive pattern matching. How do we make it error-proven and reduce the boilplate at same time is a question for the GHC folks. I don't want to take any guess here.

shineli1984:
Good work!

PR #119, opened 2019-10-02
Week 8 Ex 2

tienwei:
Thanks!


cis194/week8/tien/src/Party.hs
```diff
@@ -20,3 +20,14 @@ instance Monoid GuestList where
...
 moreFun = max
+
+-- exercie 2 --
+data Tree a =
+  Node
+    { rootLabel :: a -- label value
+    , subForest :: [Tree a] -- zero or more child trees
+    }
+
+treeFold :: b -> (a -> [b] -> b) -> Tree a -> b
```
stevemao: There's no need to have default value `b` here, since you could decide what to do with an empty list of type `[b]`

tienwei: Make sense! Thanks @stevemao :)

cis194/week8/tien/src/Party.hs
```diff
@@ -20,3 +20,14 @@ instance Monoid GuestList where
 -- 1.3 --
 moreFun :: GuestList -> GuestList -> GuestList
 moreFun = max
+
+-- exercie 2 --
+data Tree a =
+  Node
+    { rootLabel :: a -- label value
+    , subForest :: [Tree a] -- zero or more child trees
+    }
```
stevemao: You could import `Data.Tree` if you don't want to copy the code.

PR #122, opened 2019-10-02
Exercise 2: Folding with trees

stevemao:


cis194/week4/stevemao/Homework.hs
```diff
@@ -28,6 +28,27 @@ fn2 a
...
+  
+getHeight :: Tree a -> Integer
+getHeight Leaf = -1
+getHeight (Node height _ _ _) = height
+
+insertTree :: a -> Tree a -> Tree a
+insertTree a Leaf = Node 0 Leaf a Leaf
+insertTree a (Node height left b right) 
+  | getHeight left < getHeight right = Node height left' b right 
+  | otherwise = Node height' left b right'
```
shineli1984: When left height > right height, why is height always added by 1? It seems not consistent with the line above. Can we test this?

stevemao: > When left height > right height, why is height always added by 1?

It's not always + 1 because the height of right may not change.

> It seems not consistent with the line above.

Because we always insert to right first if `getHeight left == getHeight right`. So if we insert to left, the height won't change. `getHeight left > getHeight right` shouldn't be possible

stevemao: I stole jason's test file and it seems fine.

PR #133, opened 2019-10-09
Week 10

hvariant:
😺


hvariant:
it's possible to simplify a lot of the implementations using first, closing this for now

PR #135, opened 2019-10-10
Stevemao w11

stevemao:


cis194/week11/stevemao/SExpr.hs
```diff
@@ -0,0 +1,56 @@
...
+data Atom = N Integer | I Ident
+  deriving Show
+
+-- An S-expression is either an atom, or a list of S-expressions.
+data SExpr = A Atom
+           | Comb [SExpr]
+  deriving Show
+
+parseSExpr :: Parser SExpr
+parseSExpr = spaces *> noSpace <* spaces
```
shineli1984: nit:
Does `p1 *> p <* p2` mean `p` between `p1` and `p2`? if so, it seems like we can have a parser called `between`.

cis194/week11/stevemao/SExpr.hs
```diff
@@ -0,0 +1,56 @@
...
+  where noSpace = toN posInt <|> toI ident <|> char '(' *> oneOrMoreSExprs <* char ')'
```
shineli1984: nit:
With `between` parser, we can build a parser called `inBrackets` and use it here.

PR #136, opened 2019-10-10
Week10

stevemao:


cis194/week10/stevemao/AParser.hs
```diff
@@ -0,0 +1,116 @@
...
+
+first :: (a -> b) -> (a,c) -> (b,c)
+first f (a, c) = (f a, c)
+
+maybeFirst :: (a -> b) -> Maybe (a,c) -> Maybe (b,c)
+maybeFirst _ Nothing = Nothing
+maybeFirst f (Just a) = Just . first f $ a
+
+instance Functor Parser where
+  fmap f (Parser a) = Parser $ maybeFirst f . a
```
cht8687: could directly use:

Parser $ fmap (first f)  . a

cis194/week10/stevemao/AParser.hs
```diff
@@ -0,0 +1,112 @@
...
+    f xs
+      | null ns   = Nothing
+      | otherwise = Just (read ns, rest)
+      where (ns, rest) = span isDigit xs
+
+------------------------------------------------------------
+-- Your code goes below here
+------------------------------------------------------------
+
+first :: (a -> b) -> (a,c) -> (b,c)
```
shineli1984: nit
```suggestion
first :: (a -> b) -> (a ,c) -> (b, c)
```

cis194/week10/stevemao/AParser.hs
```diff
@@ -0,0 +1,112 @@
...
+abParser_ :: Parser ()
+abParser_ = void abParser
+
+spaceParser :: Parser Char
+spaceParser = Parser f 
+  where f (' ' : xs) = Just (' ', xs)
+        f _ = Nothing
+
+combiner ::() -> Integer -> Char -> Integer -> [Integer]
+combiner _ s1 _ s2 = [s1,s2]
```
shineli1984: there are `*>` `<*` we can use to get rid of this function.

cis194/week10/stevemao/AParser.hs
```diff
@@ -0,0 +1,112 @@
...
+
+sanityParser :: Parser ()
+sanityParser = Parser f
+  where f xs@[x, x', ' ', y, y'] = if isDigit x && isDigit x' && isDigit y && isDigit y'
```
shineli1984: Why do we need `sanityParser`?

stevemao: The homework is a bit ambiguous. EG: it doesn't say what happens to `intPair "123 34"`

shineli1984: > • Create a parser intPair which reads two integer values separated
> by a space and returns the integer values in a list. You should use
> the provided posInt to parse the integer values.
In the case of "123 34", both "123" and "34" are integer values.

PR #142, opened 2019-10-13
[Week7]-Task3..[Week10]: Task1, Task2

cht8687:
class Functor f where  
    fmap :: (a -> b) -> f a -> f b  

  Functor law:
    fmap id = id
    fmap (g . h) = fmap g . fmap h

Applicative Functor:

 class Functor f => Applicative f where
  pure :: a -> f a
  (<*>) :: f (a -> b) -> f a -> f b

Applicative law:
  pure id <*> x = x
  pure (g x) = pure g <*> pure x
  x <*> pure y = pure (\g -> g y) <*> x
  x <*> (y <*> z) = (pure (.) <*> x <*> y) <*> z


cis194/week7/haotian/week7/src/Scrabble.hs
```diff
@@ -0,0 +1,38 @@
...
+
+import Data.Monoid
+import Data.Char
+
+newtype Score = Score Int
+  deriving (Eq, Read, Show, Ord, Num)
+
+instance Monoid Score where
+  mempty  = Score 0
+  mappend = (+)
```
stevemao: This is duplicated definition.

cht8687: explain to me why? I am dumb. 😄 

does it because Monoid belongs to Semigroup?

stevemao: `mappend` is `(<>)` which is already defined below

cht8687: @stevemao great! thanks

PR #144, opened 2019-10-15
Week 10 Ex 3.1, 3.2

tienwei:
Thanks!


cis194/week10/tien/src/AParser.hs
```diff
@@ -69,7 +69,36 @@ instance Functor Parser where
...
+    a = formatter 'a'
+    b (_:xs) = formatter 'b' $ xs
+    b [] = Nothing
+
+-- 3.2 --
+emptyPair :: (a, b) -> ()
+emptyPair (_, _) = ()
+
+abParser_ :: Parser ()
+abParser_ = emptyPair <$> abParser
```
stevemao: You can just use `void` here.

tienwei: Thanks @stevemao. Do you have any useful link about that information? Is the function from `Data.Void`?

stevemao: https://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Monad.html#v:void

tienwei: Updated! Thanks.

cis194/week10/tien/src/AParser.hs
```diff
@@ -69,7 +70,33 @@ instance Functor Parser where
...
+pair x y = (y, x)
+
+drop1Pair :: (a1, [a2]) -> (a1, [a2])
+drop1Pair x = (fst x, drop 1 . snd $ x)
+
+formatter :: Char -> String -> Maybe (Char, [Char])
+formatter x = fmap drop1Pair . runParser (char x)
+
+abParser :: Parser (Char, Char)
+abParser = pair <$> Parser b <*> Parser a
```
stevemao: I think `char 'a'` and `char 'b'` should be enough here?

cis194/week10/tien/src/AParser.hs
```diff
@@ -69,7 +70,33 @@ instance Functor Parser where
...
+    where
+      x y = Just (a, y)
+  Parser f <*> Parser a = Parser b
     where
       b xs = pure first <*> fmap fst (f xs) <*> (a xs)
+
+-- exercise 3 --
+-- 3.1 --
+pair :: Char -> Char -> (Char, Char)
+pair x y = (y, x)
```
stevemao: This can just be `(,)`, which is an operator to create a pair

PR #147, opened 2019-10-21
Daniel M Week11

danielmcmillan:


cis194/week11/daniel-mc/SExpr.hs
```diff
@@ -0,0 +1,65 @@
...
+  deriving (Show, Eq)
+
+parseAtom :: Parser Atom
+parseAtom = N <$> posInt <|> I <$> ident
+
+parseSExprList :: Parser [SExpr]
+parseSExprList = char '(' *> oneOrMore parseSExpr <* char ')'
+
+parseSExpr :: Parser SExpr
+parseSExpr = spaces *> (A <$> parseAtom <|> Comb <$> parseSExprList) <* spaces
```
shineli1984: nit: We can possibly have a parser called `between`, `betweenBrackets`, `betweenSpaces` to reduce some duplications in the pattern of `p1 *> p <* p2`

PR #152, opened 2019-10-27
Week10/task3 4

cht8687:


cis194/week10/haotian/src/AParser.hs
```diff
@@ -105,3 +105,22 @@ instance Applicative Parser where
                           case f s of
                             Nothing -> Nothing
                             Just (r, s') -> fmap (first r) . g $ s'
+
+-- Ex. 3
+abParser :: Parser (Char, Char)
+abParser = (,) <$> char 'a' <*> char 'b'
+
+abParser_ :: Parser ()
+abParser_ = (\a b -> ()) <$> char 'a' <*> char 'b'
```
tienwei: Here as @stevemao suggested before, can use `void`. Also can just reuse `abParser`.

cis194/week10/haotian/src/AParser.hs
```diff
@@ -105,3 +105,22 @@ instance Applicative Parser where
...
+intPair = (\x _ y -> [x, y]) <$> posInt <*> char ' ' <*> posInt
+
+-- Ex. 4
+instance Alternative Parser where
+    empty = Parser $ const Nothing
+    (Parser g) <|> (Parser f) = Parser $ \s -> (g s <|> f s)
+
+intOrUppercase :: Parser ()
+intOrUppercase =
+    ((\x -> ()) <$> posInt) <|> ((\x -> ()) <$> satisfy (isUpper))
```
tienwei: Here could also apply `void` instead of `(\x -> ())`.

PR #153, opened 2019-10-27
Week11

cht8687:


cis194/week11/haotian/src/SExpr.hs
```diff
@@ -0,0 +1,60 @@
...
+
+oneOrMore :: Parser a -> Parser [a]
+oneOrMore p = (:) <$> p <*> zeroOrMore p
+
+------------------------------------------------------------
+--  2. Utilities
+------------------------------------------------------------
+
+spaces :: Parser String
+spaces = (:) <$> satisfy isSpace <*> spaces <|> pure ""
```
tienwei: Here you could probably take advantage of `zeroOrMore`.

PR #154, opened 2019-12-23
Week 7 exercise

zehua:
long overdue 😅


cis194/week7/StringBuffer.hs
```diff
@@ -9,12 +9,14 @@ import Buffer
 newtype BufferString = BufferString { getString :: String }
 
 instance Buffer BufferString where
-  toString     = getString . id
-  fromString   = id . BufferString
+  toString     = getString
+  fromString   = BufferString
```
zehua: not sure why I added `id` in #98

cis194/week7/StringBuffer.hs
```diff
@@ -9,12 +9,14 @@ import Buffer
...
   line n       = safeIndex n . lines . toString
-  replaceLine n l = fromString . unlines . uncurry replaceLine' . splitAt n . lines . toString
-      where replaceLine' pre [] = pre
-            replaceLine' pre (_:ls) = pre ++ l:ls
+  replaceLine n l
+        | n < 0     = id
+        | otherwise = fromString . unlines . uncurry replaceLine' . splitAt n . lines . toString
+          where replaceLine' pre [] = pre
+                replaceLine' pre (_:ls) = pre ++ l:ls
```
zehua: fix the bug when `n == -1`. The old version would replace first item with `l` when ` n == -1`. The correct behaviour should be to return the `Buffer` untouched.

cis194/week7/zehua/src/JoinList.hs
```diff
@@ -0,0 +1,174 @@
...
+
+
+-- ex2
+sizedTag :: (Sized m, Monoid m) => JoinList m a -> m
+sizedTag = tag
+
+getSizeFromTag :: (Sized m, Monoid m) => JoinList m a -> Int
+getSizeFromTag = getSize . size . sizedTag
+
+foldJoinList :: (Sized b, Monoid b) =>
```
zehua: the fold version seems overly complicated. I included them here just as an exercise.

cis194/week7/zehua/stack.yaml.lock
```diff
@@ -0,0 +1,12 @@
+# This file was autogenerated by Stack.
```
zehua: do we usually check in the lock file?

cis194/week7/zehua/test/JoinListBufferSpec.hs
```diff
@@ -0,0 +1,95 @@
...
+            (ass 23 2
+              (sss 9 1 "klm")
+              (sss 14 1 "opq"))
+
+infix 1 `shouldBeJLEqualTo`
+shouldBeJLEqualTo :: JoinList (Score, Size) String -> JoinList (Score, Size) String -> Expectation
+actual `shouldBeJLEqualTo` expected = do
+  toLines actual `shouldBe` toLines expected
+  getJoinListScore actual `shouldBe` getJoinListScore expected
+  getJoinListSize actual `shouldBe` getJoinListSize expected
```
zehua: not sure if this is the right way to build helper assert functions in Haskell

shineli1984: they are just normal functions so this is fine.

cis194/week7/zehua/stack.yaml.lock
shineli1984: https://github.com/commercialhaskell/stack/issues/4795

PR #155, opened 2019-12-29
Zehua Week 8

zehua:
`stack run` to run the main program.

Output:
```
Total fun: 21669
Adam Debergues
Adam Donalon
Adelaide Ahrens

.
.
.

Vitta Dennemeyer
Walker Burgeat
Walter Bradford
Wilhelmine Backes
Wilhelmine Blazi
William Bezdeck
```


cis194/week8/zehua/src/Party.hs
```diff
@@ -0,0 +1,52 @@
...
+import           Data.List (sort)
+import           Data.Tree (Tree (Node, rootLabel, subForest))
+import           Employee
+
+-- ex1
+glCons :: Employee -> GuestList -> GuestList
+glCons e gl = GL (e:el) (f + glf)
+  where
+    Emp { empFun = f } = e
+    GL el glf = gl
```
zehua: not sure which one is better: inline pattern matching or doing it in `where`.

shineli1984: Inline pattern match is used more when pattern match on constructors

PR #156, opened 2019-12-30
Zehua Week10 hw

zehua:


cis194/week10/zehua/src/AParser.hs
```diff
@@ -0,0 +1,157 @@
...
+    where
+      pc s = runParser pa s >>= fb
+      fb (fa, sa) = runParser (fmap fa pb) sa
+  -}
+  {- using Parser Functor and Maybe Monad with do notation, probably the simplest
+  pa <*> pb = Parser pc
+    where
+      pc s = do
+        (fa, sa) <- runParser pa s
+        runParser (fmap fa pb) sa
```
zehua: is there a way to eliminate `s`?

cis194/week10/zehua/src/AParser.hs
```diff
@@ -0,0 +1,157 @@
...
+  -}
+  {- using Parser Functor and manual Maybe mapping / flattening
+  pa <*> pb = Parser pc
+    where
+      pc s = oa
+        where
+          ra = runParser pa s
+          oa = case ra of
+                 Nothing -> Nothing
+                 Just (fa, sa) -> runParser (fmap fa pb) sa
```
zehua: not sure if it's common to do it in this type of nested `where` structure.

cis194/week10/zehua/src/AParser.hs
```diff
@@ -0,0 +1,157 @@
...
+-- Your code goes below here
+------------------------------------------------------------
+
+-- ex1
+first :: (a -> b) -> (a, c) -> (b, c)
+first f (a, c) = (f a, c)
+
+instance Functor Parser where
+  -- fmap :: (a -> b) -> Parser a -> Parser b
+  -- fmap :: (a -> b) -> (String -> Maybe (a, String)) -> (String -> Maybe (b, String))
```
zehua: this is technically not correct. just expanding it out to help understand the type

cis194/week10/zehua/src/AParser.hs
shineli1984: There always is a way to eliminate all the argument. But I don’t think it’ll be pretty here

cis194/week10/zehua/src/AParser.hs
shineli1984: Nested where should be fine. But in this case we can inline oa

PR #158, opened 2019-12-31
Zehua Week12

zehua:
technically still in 2019 (Beijing time) 😅 

didn't have enough time to do the optional ex 5 :|

to test the program, supply two numbers to the command line arguments. If not supplied, it will ask via stdin.

```
Usage: stack run [<attackers> <defenders>]

$ stack run 100 100
Probability of winning is 0.804

$ stack run
Enter number of attackers:
100
Enter number of defenders:
100
Probability of winning is 0.8
```


cis194/week12/zehua/app/Main.hs
```diff
@@ -0,0 +1,27 @@
...
+  a <- readLn
+  putStrLn "Enter number of defenders: "
+  d <- readLn
+  return (a, d)
+
+main :: IO ()
+main = do
+  args <- getArgs
+  -- get from command line args first, if not from stdin
+  (a, d) <- fromMaybe getParamsFromStdin $ getParamsFromArgs args
```
zehua: laziness FTW!

cis194/week12/zehua/src/Risk.hs
```diff
@@ -0,0 +1,103 @@
...
+first :: (a -> b) -> (a, c) -> (b, c)
+first f (a, c) = (f a, c)
+
+instance Random DieValue where
+  random           = first DV . randomR (1,6)
+  randomR (low,hi) = first DV . randomR (max 1 (unDV low), min 6 (unDV hi))
+
+-- make gen generic to help testing
+-- die :: Rand StdGen DieValue
+die :: RandomGen g => Rand g DieValue
```
zehua: changed the signature slightly to support testing via mocked `RandomGen`

cis194/week12/zehua/test/RiskSpec.hs
```diff
@@ -0,0 +1,220 @@
...
+    it "works for dieNSorted via MockRandomGen" $ do
+      (flip evalRand) (mkMockRandomGen [1,2,3,4,5,6]) $ do
+        r1 <- dieNSorted 1
+        r2 <- dieNSorted 3
+        r3 <- dieNSorted 2
+        return (\() -> do
+          r1 `shouldBe` [1]
+          r2 `shouldBe` [4,3,2]
+          r3 `shouldBe` [6,5]
+          )
```
zehua: really could not get `Rand g a` and `Expectation` to work together. So this is the hacky way in which I managed to make the compiler agree to work 😅 

The main idea is that since I could not get the two monads to fuse together, I just run the `Rand g a` one first to produce the assertions and execution the assertions in the `IO` / `Expectation` world.

